Issue #85, opened 2019-10-12
Warning is printed twice


linsui:
I got this message.

```
WARNING: using legacy espanso config location in $HOME/.espanso is DEPRECATED
Starting from espanso v0.3.0, espanso config location is changed.
Please check out the documentation to find out more: https://espanso.org/docs/configuration/
WARNING: using legacy espanso config location in $HOME/.espanso is DEPRECATED
Starting from espanso v0.3.0, espanso config location is changed.
Please check out the documentation to find out more: https://espanso.org/docs/configuration/
```

federico-terzi:
Hi @linsui,
Thanks for the feedback. I will try to solve the bug in the next release. 
Regards

Boxfort:
I've proposed a fix for this in PR #87

PR #87, opened 2019-10-15
only call get_config_dir() once

Boxfort:
This is a fix for issue #85.

Stores the output of 'get_config_dir' so that the `eprintln!` statements within are not called multiple times.


federico-terzi:
Hi,
Thank you for your help! Unfortunately the problem is more problematic than this, because the multiple prints are triggered also for other subcommands, in places where is not so easy to solve. It would probably take `lazy_static` or some kind of Singleton to fix. I will take a look at it in a few days using your feedbacks :)

PR #136, opened 2019-12-11
preserve the current clipboard when using the Clipboard backend.

matt-h:
Store the current clipboard and restore it after the expansion has been pasted.
This prevents your current clipboard from being lost if using the Clipboard backend.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -205,8 +205,10 @@ impl <'a, S: KeyboardManager, C: ClipboardManager, M: ConfigManager<'a>, U: UIMa
                         }
                     },
                     BackendType::Clipboard => {
+                        let previous_clipboard_content = self.clipboard_manager.get_clipboard().unwrap_or(String::from(""));
                         self.clipboard_manager.set_clipboard(&target_string);
                         self.keyboard_manager.trigger_paste(&config.paste_shortcut);
+                        self.clipboard_manager.set_clipboard(&previous_clipboard_content);
                     },
                 }
 


federico-terzi:
Hi @matt-h,

Thank you for your help, I've been thinking about this feature for a while now. The only problem with your solution is that clipboard content type can be different from text, and in that case, this method will not work. For example, if the clipboard contained an image, `get_clipboard` does not handle it.

That said, I'll try to include your pull request in the next release, adding a configuration option to allow the user to enable/disable it in case.

I'll keep you updated.

Thanks for the help :)

matt-h:
Yeah, I thought about that as well. I wasn't sure how I could include it so it could preserve the clipboard for other things but text. The way I did it, if it is something besides text it will just clear out your clipboard after the expansion. Not ideal, but as it is now if the expansion is saved in the clipboard then your saved clipboard is lost anyway.

I agree that it should be configurable, perhaps in the same spot you set the backend so it can be done per command if you only want the expansion saved in the clipboard for some commands.

Ideally if would be on by default that it preserves the clipboard unless specified not too. My guess would be it is jarring that the expansion would override you clipboard for new users to the tool like it was for me.

Maybe keep it an option that is off by default in a minor release, but then in the next major release it can be switched to enabled by default.

federico-terzi:
Hey @matt-h,

> Yeah, I thought about that as well. I wasn't sure how I could include it so it could preserve the clipboard for other things but text. The way I did it, if it is something besides text it will just clear out your clipboard after the expansion. Not ideal, but as it is now if the expansion is saved in the clipboard then your saved clipboard is lost anyway.

I agree, at least there is a chance to save the current content of the clipboard if it is text.

> I agree that it should be configurable, perhaps in the same spot you set the backend so it can be done per command if you only want the expansion saved in the clipboard for some commands.

Yep, that's exactly how I thought it.

> Ideally if would be on by default that it preserves the clipboard unless specified not too. My guess would be it is jarring that the expansion would override you clipboard for new users to the tool like it was for me.

I completely understand your point, it makes sense. Originally, the `Clipboard` backend wasn't the default one on Linux and therefore this wasn't a problem. I had to make this change though because the other one was pretty buggy and not in a way that I could solve.

> Maybe keep it an option that is off by default in a minor release, but then in the next major release it can be switched to enabled by default.

This seems like the best approach so that we get some testing before the general release.

Thanks for your help, I'll keep you updated

federico-terzi:
Hey @matt-h,

Your contribution is currently making its way to version 0.4.1 that will be released shortly :)
I made a few minor modifications to include also image matches.

You are now officially an espanso's contributor! I listed you Readme

Thanks for your help :)

PR #170, opened 2020-01-31
Add native Windows Notifications

### LINKED ISSUE #169: Native Notification on Windows
Hi,

The current notification looks a little out of place on Windows:
image

It would be nice to get a toast notification instead, which is possible using [winrt](https://github.com/contextfree/winrt-rust/blob/master/examples/toast_notify.rs), for example.

If this is a lower priority for you, I'd love to try setting it up with some guidance.

Great project!

LiHRaM:
closes #169

***Couple of things***: In order for native notifications to register as coming from espanso, it needs to register an [AppUserModelID](https://msdn.microsoft.com/en-us/library/windows/desktop/dd378459(v=vs.85).aspx) during installation.
Currently, [winrt_notification](https://allenbenz.github.io/winrt-notification/0_2_0/winrt_notification/struct.Toast.html) provides a workaround by just using Windows PowerShell's ID.

Also, I had to do a little hacking with paths, so that it correctly uses the icon both when developing on Windows, and when the app's been installed.

Screenshot:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,6 +78,11 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "bitflags"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
@@ -394,6 +399,7 @@ dependencies = [
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "walkdir 2.2.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "widestring 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winrt-notification 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "zip 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -970,6 +976,11 @@ dependencies = [
 
 [[package]]
 name = "quote"
+version = "0.3.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "quote"
 version = "0.6.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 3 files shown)


LiHRaM:
The Windows CI is failing, but it doesn't look like the failure is caused by any changes I made?

cc @federico-terzi

federico-terzi:
Hey @LiHRaM,
That notification look fantastic, thank you for your efforts!
Unfortunately, I won't be able to access my computer until tomorrow, but from a preliminary analysis it seems like the CI pipeline broke in the rust installation stage, so nothing related to your commit. Those images are changed rather frequently so it happens sometimes, tomorrow I'll try fixing it up.

Regarding the icon path, I'm going to take a closer look tomorrow, but there could be another way to solve the problem.

Regarding the addition of AppId during installation, we'll need to find a way to integrate it with InnoSetup, the packaging system currently used by espanso. But I don't think it will be much of a problem.

Thank you very much for your help :)

LiHRaM:
No problem! Thanks for maintaining this project, I love it!

federico-terzi:
As a note to myself, this seems to solve the issue: https://github.com/rust-lang/rustup/issues/2203

federico-terzi:
Hey @LiHRaM,
I finally had a chance to try your request. First of all, great work, the integration looks super good!

There is a problem though that I'd like to address before merging it into espanso.
I often toggle espanso on/off, but when I do that with the new native notification, all the toast messages get queued, so that instead of updating the notification in realtime, it get delayed by a few seconds,. This becomes annoying when toggling many times in a row, as the user keeps receiving notifications for like 30 seconds.

Do you have any idea on how to solve it?

Thanks for your help :)

LiHRaM:
Hey @federico-terzi,
No, I'm not aware of a way to dequeue older notifications in favor of 
the latest.
In fact, I'm pretty sure the Windows team considers this a feature.
The easiest thing to do would probably be to consider only showing 
notifications for startup/shutdown,
and visually indicate that espanso is disabled/enabled in a different 
way.

In any case, there's no option for it in the bindings that I chose, but 
maybe there is in the more raw bindings they're built on. (winrt vs 
winrt-notifications)

federico-terzi:
Hey @LiHRaM,
Thank you for the suggestion, I'll try taking a look at the raw notification API and see if there's an option. 
In fact your pull request was a great help in that sense as it directed me to the right documentation resources, that I couldn't find before.
I'll keep you updated
Thanks :)

PR #179, opened 2020-02-13
Format using 'cargo fmt'

Atul9:


federico-terzi:
Hey @Atul9,
I really appreciate your help, unfortunately this pull request is changing too many files at once to merge it into the current version of espanso.
I'm planning to convert espanso to standard rust formatting in the next major version (0.6.0) as its effects can be problematic.

Cheers :)

PR #194, opened 2020-03-05
Add support for kitty terminal

miterion:
Currently replacement does not work in kitty, since it needs the same special treatment as tilix. This pr simply adds the necessary workaround.


federico-terzi:
Hey @miterion,
Thank you very much for your help!

Before merging, I'd like to make sure it works correctly. Have you tested it?

Cheers :)

miterion:
Yeah, I did a cargo build after which it worked in daemon mode. It did work after that (only used a few test commands though).

federico-terzi:
> Yeah, I did a cargo build after which it worked in daemon mode. It did work after that (only used a few test commands though).

Seems perfect! I'll merge it into `dev` so that it gets shipped in the next release!

Thanks for your help :)

PR #359, opened 2020-07-10
Adds a debug option for espanso when executing shell commands for greater context.

deckarep:
In reference to the following: https://github.com/federico-terzi/espanso/issues/355

I've added a `debug` flag for shell commands to utilize.

- When not provided debug defaults to `false`.
- When non-parseable debug defaults to `false`.
- When provided, error log is thrown which captures: cmd string, exit_status, stderr, stdout in a single log-line.
- I can attempt to add a relevant unit-test if this looks like a reasonable patch.


src/extension/shell.rs
```diff
@@ -171,6 +171,19 @@ impl super::Extension for ShellExtension {
...
+                let debug_opt = params.get(&Value::from("debug"));
+                let with_debug = if let Some(value) = debug_opt {
+                    let val = value.as_bool();
+                    val.unwrap_or(false)
+                }else{
+                    false
+                };
+                
+                if with_debug {
+                    error!("debug for shell cmd '{}', exit_status '{}', stdout '{}', stderr '{}'", cmd, output.status, output_str, error_str);
```
federico-terzi, line 184: Could you please use info!() instead of `error`?

deckarep: Yes no problem, and thanks for clarifying on merging into `dev` first.

federico-terzi:
Hey @deckarep,
I'm currently away and I'll get back home by Monday, I'll make sure to review the changes and merge it into dev next week.
Thanks again for your help :)

deckarep:
Sounds good @federico-terzi!

One question would you also be interested in PRs that provide optimizations both big and small? Some examples that come to mind I see are PRs where owned strings are allocated instead of using references. Of course bigger optimizations could be present in hot paths that might be more impactful.

federico-terzi:
> One question would you also be interested in PRs that provide optimizations both big and small? Some examples that come to mind I see are PRs where owned strings are allocated instead of using references. Of course, bigger optimizations could be present in hot paths that might be more impactful.

@deckarep In general, I'd like to discuss all the PRs that get proposed as I really want to avoid wasting people time :) 

I originally started espanso to learn Rust, so some of the early work was definitely not idiomatic. If I had the time, I would completely re-write it from scratch now that I have much better knowledge, but that would be mostly for aesthetics as espanso is already fast enough for most use-cases.

federico-terzi:
Merged, thanks for the help :)

PR #368, opened 2020-07-13
Restore clipboard to its pre-trigger state before evaluating passive match

### LINKED ISSUE #365: clipboard contents getting overwritten during passive match (xclip)
Arch Linux, espanso 0.6.3, GNU `sed`

The goal is to indent the clipboard contents, line by line, with a given number of spaces. I have a similar active match that always does 4 spaces, but want to add one that prepends an arbitrary number of spaces, and the best I could think of so far was cramming a looped `printf` in there.

```yaml
backend: Clipboard
fast_inject: false
enable_passive: true
passive_key: CTRL

matches:

  - trigger: :in
    replace: "{{indented}}"
    passive_only: true
    vars:
      - name: indented
        type: shell
        params:
          cmd: sed "s/^/$(for i in {1..$0}; do printf "%s" " "; done)/g" <<<"$(xclip -sel clip -o)"
          trim: false
```

```console
$ printf '%s\n' line1 line2 | xclip -sel clip
$ xclip -sel clip -o
line1
line2
```

Before and after:
```
:in/2/

  :in/2/

```

```console
$ xclip -sel clip -o
  :in/2/
```

I don't have any syncing going on between `XA_PRIMARY` and `XA_CLIPBOARD`; without running espanso, I can select text after text, but the output of `xclip -sel clip -o` will not change. It seems during a passive replacement, espanso writes over the contents of `XA_CLIPBOARD`.

AndydeCleyre:
Fixes #365

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -366,6 +366,7 @@ impl<
                 info!("Avoiding passive expansion, as the user didn't select anything");
             } else {
                 if let Some(previous_content) = previous_clipboard {
+                    self.clipboard_manager.set_clipboard(&previous_content);
                     // Because of issue #213, we need to make sure the user selected something.
                     if clipboard == previous_content {
                         info!("Avoiding passive expansion, as the user didn't select anything");


AndydeCleyre:
Obsoleted by merged (:fireworks:) #373

PR #373, opened 2020-07-15
Tighten passive expansion process

### LINKED ISSUE #365: clipboard contents getting overwritten during passive match (xclip)

### LINKED ISSUE #372: Passive expansion does not happen if the selection is already in the clipboard when triggering
### Reproduction

- Select text for passive replacement
- Copy to clipboard (e.g. as a backup of the content before seeing how it's replaced)
- Trigger passive replacement

Replacement does not occur, as espanso has guessed that there is no text currently selected to replace.

---

### Background

This was introduced as a fix for #213, with 4d91085. The requirement there is to prevent a passive expansion if no text is actually selected at time of trigger.

That fix takes the following approach, as I understand it:

1. backup current clipboard to `previous_clipboard`
2. trigger keyboard copy for current selection into clipboard
3. read clipboard into variable, let's call this `to_expand`
4. if `to_expand` is `None`/`""`/whitespace, abort; otherwise, continue
5. if `previous_clipboard` is `None`, abort; otherwise, continue
6. if `to_expand` equals `previous_clipboard`, abort; otherwise, continue
7. render expansion
8. set clipboard to expansion
9. trigger keyboard paste

---

#### Primary

- I don't know the purpose of `(5)` (if I'm reading the code right, anyway)
- `(6)` is the direct cause of this issue
- We can set the clipboard to `""` between `(1)` and `(2)`, and in `(6)` compare `to_expand` to `""`/whitespace instead of `previous_clipboard`

#### Secondary

- Upon abortion at any stage, it would be nice to restore `previous_clipboard` to the clipboard
- We should also restore `previous_clipboard` to the clipboard before rendering (#365, #368)
- To do both of those, we can add two restore calls: one right after `(3)`, and another after `(9)` 

---

#### Intentional Clipboard Destruction

Before looking into this, I didn't realize there was a `preserve_clipboard` option (which seems to default to `true`), so I haven't accounted for folks with that set to `false`. But it may be already handled outside the scope of this function anyway :shrug:

#### Oof, Images

@federico-terzi [said](https://github.com/federico-terzi/espanso/issues/213#issuecomment-658275372):

> The only problem I see is that if the user has currently an image copied in the clipboard, this process won't be able to recover it.

- The current `ClipboardManager` only gets `str`/`None` out of the clipboard, I see. I would personally take that tradeoff over the current one.
- If you want to make image clipboard content handling more robust, FWIW:

```console
$ xclip -sel clip <<<"content"
$ xclip -sel clip -o -t TARGETS
TARGETS
UTF8_STRING
```
After copying some text from a browser:
```console
$ xclip -sel clip -o -t TARGETS
TIMESTAMP
TARGETS
MULTIPLE
SAVE_TARGETS
UTF8_STRING
COMPOUND_TEXT
TEXT
STRING
text/plain;charset=utf-8
text/plain
```
```console
$ xclip -sel clip -t image/png < mylatest.png
$ xclip -sel clip -o -t TARGETS
TARGETS
image/png
```
After copying image data using my desktop environment:
```console
xclip -sel clip -o -t TARGETS
application/x-qt-image
image/png
image/bmp
image/bw
image/cur
image/eps
image/epsf
image/epsi
image/icns
image/ico
image/jp2
image/jpeg
image/jpg
image/pbm
BITMAP
image/pcx
image/pgm
image/pic
image/ppm
PIXMAP
image/rgb
image/rgba
image/sgi
image/tga
image/tif
image/tiff
image/wbmp
image/webp
image/xbm
image/xpm
TARGETS
MULTIPLE
TIMESTAMP
SAVE_TARGETS
$ xclip -sel clip -o -t image/jpg >> qimg1
$ xclip -sel clip -o -t image/png >> qimg2
$ file qimg*
qimg1: JPEG image data, JFIF standard 1.01, resolution (DPI), density 96x96, segment length 16, baseline, precision 8, 267x159, components 3
qimg2: PNG image data, 267 x 159, 8-bit/color RGB, non-interlaced
```

AndydeCleyre:
- Don't abort if original clipboard is None (why should we?)
- Don't abort if original clipboard has the same text as that selected
- If original clipboard has text, restore that text before render
- If original clipboard has text, restore that text in case of aborted expansion

Note: will unfortunately remove non-text data (image) from clipboard if no text
is selected when passive expansion is triggered

Fixes #372

Fixes #365 (supersedes #368)

I left the hardcoded delays intact, though it's not clear exactly when they're necessary.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -95,6 +95,9 @@ fn default_passive_arg_delimiter() -> char {
 fn default_passive_arg_escape() -> char {
     '\\'
 }
+fn default_passive_delay() -> u64 {
+    100
+}
 fn default_passive_key() -> KeyModifier {
     KeyModifier::OFF
 }
@@ -205,6 +208,9 @@ pub struct Configs {
 
     #[serde(default = "default_passive_key")]
     pub passive_key: KeyModifier,
+
+    #[serde(default = "default_passive_delay")]
+    pub passive_delay: u64,
 
     #[serde(default = "default_enable_passive")]
     pub enable_passive: bool,
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -346,10 +346,13 @@ impl<
         // In order to avoid pasting previous clipboard contents, we need to check if
         // a new clipboard was effectively copied.
         // See issue: https://github.com/federico-terzi/espanso/issues/213
-        let previous_clipboard = self.clipboard_manager.get_clipboard();
+        let previous_clipboard = self.clipboard_manager.get_clipboard().unwrap_or_default();
 
         // Sleep for a while, giving time to effectively copy the text
         std::thread::sleep(std::time::Duration::from_millis(100)); // TODO: avoid hardcoding
+
+        // Clear the clipboard, for new-content detection later
+        self.clipboard_manager.set_clipboard("");
 
         // Trigger a copy shortcut to transfer the content of the selection to the clipboard
         self.keyboard_manager.trigger_copy(&config);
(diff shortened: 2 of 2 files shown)


federico-terzi:
Thank you @AndydeCleyre, I tried your PR and is mostly working as expected, fixing those errors. That said, I experienced some edge-cases on Windows, therefore I added a couple of delay calls (they are needed to make sure the target app has the time to save/copy from the clipboard).

Could you please review #375 (that is just a slight modification of this one) and see if that works as expected in your case? If so, we can then proceed with the merge.

Cheers :)

AndydeCleyre:
I'm actually having intermittent problems with both of these today, I'll have to look closely tonight. With my passive indenting match, I'm often getting "Passive mode activated" but it looks like it's pasting the original selection, for no net change.

Sorry, I can't get back to this for ~6-7hrs

Thank you very much.

AndydeCleyre:
It looks like you removed the clipboard restoration before the render, which would re-introduce #365

AndydeCleyre:
- I think my "intermittent problems" may have been my own mistyping of the trigger.
- The current state of this PR tests well for me against the relevant issues. It's been rebased onto a more recent dev, and merged your #375 version in; the only relevant change after your version is the latest commit, which restores the clipboard before the render, to re-fix #365.
- I don't know if another delay would be helpful, please have a look.
- Feel free to take this / squash this / have me squash this / cherry pick into yours / manually add that extra restore to yours / whatever, but this state seems good to me.
- Thanks!

federico-terzi:
> * I think my "intermittent problems" may have been my own mistyping of the trigger.
> * The current state of this PR tests well for me against the relevant issues. It's been rebased onto a more recent dev, and merged your #375 version in; the only relevant change after your version is the latest commit, which restores the clipboard before the render, to re-fix #365.
> * I don't know if another delay would be helpful, please have a look.
> * Feel free to take this / squash this / have me squash this / cherry pick into yours / manually add that extra restore to yours / whatever, but this state seems good to me.
> * Thanks!

I removed the clipboard restoration at the beginning at it seemed redundant, as we restore it anyway here: https://github.com/federico-terzi/espanso/pull/375/files#diff-8fafaf9ae6308a3c91fbf941e818c0bcR394

Do you see any problem with this approach?

Thanks for your help :)

federico-terzi:
@AndydeCleyre nevermind, I understand now! It's been a long day :P 
I agree that your solution is correct, let's proceed with this one

federico-terzi:
Merged! Thanks again for your help :) We can close #368 right?

AndydeCleyre:
Thanks! Yup, all set.

Issue #357, opened 2020-07-09
Help Wanted: Documentation


federico-terzi:
**Most of the things have been sorted out at the moment. If you want to help, you can proof-read the docs or wait until other entries make their way into the list** :)

Hey guys,

I started to notice that the documentation is becoming a bit stale, especially considering the number of features/changes that have been introduced in each release.

If any of you is willing to help me update the docs, that would be super helpful to the project :)

## Things that are missing in the docs

I always take some time to write a detailed change log in each version, so that could be a good starting point to see what changed/which are the features to document.

This is a rough list of the things that are missing:

- [x] Mac manual installation instructions (see: https://github.com/federico-terzi/espanso/issues/210#issuecomment-608617197)
- [x] At the end of the [Getting Started](https://espanso.org/docs/get-started/#useful-shortcuts) page, it might be a good idea to add a paragraph explaining how to change the `toggle_key`, as I receive a lot of requests about it (here are the possible values: https://www.reddit.com/r/espanso/comments/gemz9d/disable_or_change_double_alt_disable_function/fpodyxb?utm_source=share&utm_medium=web2x)
- [ ] Add a "Miscellaneous" section to the docs, to include useful configurations/scripts. In this section, I would add the [Rofi Integration](https://www.reddit.com/r/espanso/comments/hdzcjc/the_upcoming_version_will_make_rofi_integration/fvo3lhg?utm_source=share&utm_medium=web2x) and the [Script to convert MS dictionaries to espanso matches](https://github.com/federico-terzi/espanso/issues/337).
- [x] Remove `espanso-bin` from the list of AUR packages:  https://espanso.org/install/linux/#installing-on-manjaro--arch (until we sort out why it doesn't work reliably, we should remove it)
- [x] Add a "CLI" section to the docs, to include the new CLI options: https://github.com/federico-terzi/espanso/releases/tag/v0.6.3
- [x] Changes to the "Configuration" page:
  - [x] Explain the `auto_restart` feature: https://github.com/federico-terzi/espanso/releases/tag/v0.6.2 (which is now on by default)
  - [x] Explaining the `fast_inject` option: https://github.com/federico-terzi/espanso/releases/tag/v0.5.5 (which is now on by default)
  - [x] Document the hide icon and notifications options: https://github.com/federico-terzi/espanso/releases/tag/v0.5.5
[...]

PR #400, opened 2020-08-08
Remove espanso-bin from the list of AUR packages

bradyjoslin:
Remove espanso-bin from the list of AUR packages

requested in #357


federico-terzi:
Thanks :)

PR #421, opened 2020-08-26
Add list control to example of verbose syntax

jspaetzel:
Having the shorthand examples earlier in this doc threw me off when I wanted to do something more complex with lists. May be helpful to have an example with the other control types to make it a bit clearer when getting to advanced stuff. I think it's also useful to demo here the form with different text from what gets replaced.


federico-terzi:
@jspaetzel Seems great! Thank you for the help :)

PR #467, opened 2020-09-24
Note to avoid conflict with macOS text replacement

swrobel:
I've tried many different characters, and the only one that I found to work was /, but I'm open to other suggestions for how to document this!

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -141,6 +141,10 @@ espanso restart
 
 Now try to type `:br` anywhere. If you did everything correctly, you should see `Best Regards` appear!
 
+##### macOS remarks
+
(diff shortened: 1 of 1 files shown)


federico-terzi:
Hey @swrobel,

Thanks for your help :) I'll gladly merge this PR into the docs, but before that, I want to make sure I properly understand the problem.

> I've tried many different characters, and the only one that I found to work was /, but I'm open to other suggestions for how to document this!

So are you trying to use both the macOS built-in text expansion and espanso simultaneously, right? (may I ask why?)

And in that case, if you prefix the trigger with `/`, only espanso will trigger, am I right?

swrobel:
> So are you trying to use both the macOS built-in text expansion and espanso simultaneously, right? (may I ask why?)

I'm not really trying to, it's just impossible to disable it on macOS without deleting my replacements, which will also delete them from my iDevices, where the built-in text replacement is the only option.

> And in that case, if you prefix the trigger with /, only espanso will trigger, am I right?

Correct!

federico-terzi:
I see, thank you for the explanation!

>  it's just impossible to disable it on macOS without deleting my replacements,

That's a shame... unfortunately, as you said, there is no easy way to enable text expansion on iOS except the built-in one. Actually, there would be an option, building a custom keyboard with support for espanso, but the task is definitely out of scope given our current resources. Maybe in the future :)

PR #476, opened 2020-10-01
Fix build on mingw64 gcc compiler

mattn:


build.rs
```diff
@@ -28,6 +28,8 @@ fn get_config() -> PathBuf {
 fn print_config() {
     println!("cargo:rustc-link-lib=static=winbridge");
     println!("cargo:rustc-link-lib=dylib=user32");
+    println!("cargo:rustc-link-lib=dylib=gdiplus");
+    println!("cargo:rustc-link-lib=dylib=stdc++");
```
mattn, line 32: please wait merge. I don't understand how to make condition for the gcc compiler.

federico-terzi:
Hey @mattn,

Thanks for your contribution! Have you personally tested that espanso compiles and works correctly with the MinGW toolchain with these changes?

If you confirm I'll gladly merge your contribution :)

By the way, I've rebased the PR to target the `dev` branch, so that it will get merged with the rest of the changes in the next release.

Cheers :)

mattn:
> Have you personally tested that espanso compiles and works correctly with the MinGW toolchain with these changes?

Yes, it is.

BTW, Could you please tell me easy way to expand current location?

mattn:
```yaml
 - trigger: ":shell"
    replace: "{{output}}"
    vars:
      - name: output
        type: shell
        params:
          cmd: "pwd"
```
For example, pwd is spawned as child process of espanso. So this is not right PATH of the shell.

federico-terzi:
@mattn unfortunately, that's not trivial because espanso spawns a new shell to execute that command, which is completely unrelated to the one you are using.
That's a source of confusion for many users, but the `shell` extension is not related in any way with the current terminal you are using. In fact, it's built to execute shell commands (and get their output) in every application.

If you are using Linux or macOS, that could be possible by setting up a shell hook to write the current location on a file ([example with ZSH](https://github.com/federico-terzi/linuxconfig/blob/master/config/zshrc#L100-L102)), and then using an espanso match to paste the content of that file:

```yaml
 - trigger: ":shell"
    replace: "{{output}}"
    vars:
      - name: output
        type: shell
        params:
          cmd: "cat /tmp/lastcmddir"
```

Cheers :)

Issue #548, opened 2020-12-17
Prevent multiple espanso processes from being run after auto-reload


mamichels:
Hi there!
I was working on the config when i noticed my matched get inserted multiple times. Seems like something does now work properly with re auto reload:
![image](https://user-images.githubusercontent.com/48212402/102471634-bcde1180-4055-11eb-9407-b66b8395519e.png)

As a note, i wasn't able to close any of those through their context menues and had to use a kill command on all of them.

Obviously running it on Windows 10.

federico-terzi:
Hey @mamichels,

Thanks for reporting this issue! Indeed, this is a bug, I'll investigate and hopefully fix it in one of the next few releases :)
The most likely cause is a series of fast rewrites of the configuration files, do you have any idea of what could cause that in your system? (maybe a synchronization program?)

Cheers :)

mamichels:
Hi @federico-terzi,

tried to reproduce it without success, but this is what i done yesterday. Maybe its a clue, maybe not.
 If i want to call the config via `espanso edit` it has some issues calling notepad for whatever reason.
As shown in the terminal below, the path to the default.yml is completly valid, also the notpad.exe is correclty located. If I try it on a different program in the same terminal e.g. `npm config edit` its working fine.

![image](https://user-images.githubusercontent.com/48212402/102591650-acd83780-4112-11eb-8cb2-e76ffc059680.png)

A side note on the bugs behaviour, while there were like 5 processes only 3 of them replaced the tags. When i was using it in MS Teams it also instantly submitted the message after replacing those tags. Probably a whole other story but it may helps.

Thanks for caring!

federico-terzi:
Hey @mamichels,

That's a pretty strange behavior! I will need to do some more testing before accepting the PR, as for me the path was working... Are you using CMD or Powershell inside the Terminal?

Cheers :)

mamichels:
Hi @federico-terzi,

thats for sure on the testing part.
The issue as shown in the picture does only occure on the git bash on windows. cmd and powershell just work fine.

federico-terzi:
> does only occure on the git bash on windows. cmd and powershell just work fine.

@mamichels I see, that might explain this behavior, as perhaps bash does not understand backslashes on file paths...

fstamour:
As stated in #602 I had the exact same issue after creating a bunch of files with a script.

<hr>

[...]

PR #549, opened 2020-12-18
Replace \ with / in default_editor for windows

mamichels:
This is in regard to the minor issue mentioned in #548 
When using a bash terminal on windows there are issues with the escape character `\` which results in `notepad.exe` not being located properly.
A normal / will work just fine in a bash, cmd or powershell as tested in the screenshots below. 
image
image
image

@federico-terzi, currently i am not able to test this change in the context of espanso. If you have a small "how to build and run for dummys" i'll be happy to do so.

### DIFF
--- a/src/edit.rs
+++ b/src/edit.rs
@@ -29,7 +29,7 @@ fn default_editor() -> String {
 }
 #[cfg(target_os = "windows")]
 fn default_editor() -> String {
-    "C:\\Windows\\System32\\notepad.exe".to_owned()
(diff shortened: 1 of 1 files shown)


AucaCoyan:
Hi!
I tested the bug and I work correctly on my PC. I'm closing this PR fundamentally because it's old, but if you see some bug, don't hesitate to file an issue!
We are also gathering in discord here: https://discord.gg/DFcCNDg7bB
See you around, and sorry for the delayed response 🕐 🤣

Issue #594, opened 2021-02-15
2021 Refactor Tracking Issue


federico-terzi:
After 1.5 years since the start of this journey, it's clear how the current codebase is starting to show its limits. Espanso was my first Rust project and some of my early design choices (both caused by my lack of experience with the language and limited domain knowledge) were problematic to adapt to the new project requirements. So... It's finally time to refactor!

I'm planning to release a new major version (2.0.0), which means, *some* things might not be backward compatible for the sake of a cleaner/more robust implementation (although I'll try my best to maintain compatibility as much as possible). All development is happening in the `dev1.x` branch.

In this version, the project will heavily rely on cargo's workspaces to keep it modular and maintainable. This should reduce coupling and make it much easier to add/remove features in the future.

### Features
- [x] Implement new config handling logic to separate matches definition and app-specific configuration.
- [x] Implement the "legacy" config loader, to make espanso backward compatible with the old config format during the transition phase
- [x] Migration tool
- [x] Built-in patch system #310 
- [x] Search bar #420 
- [x] Experimental Regex triggers (with named groups to support arguments). 
- [X] Wayland support #287
- [X] Use native notifications on Windows
- [x] Module that handles the legacy/new directory mapping and migration. #611
- [x] Convert the macOS version into an App Bundle
- [x] Webview window to guide the Accessibility process on macOS
- [x] Implement new package system
- [x] Port espanso hub packages to new repository and format: https://github.com/espanso/hub

### Architectural changes
- [X] Extract the platform-specific application event-loop into a separate `ui` module
- [X] Extract the native input logic into a separate `detect` module
- [X] Extract the text injection logic into a separate `inject` module
- [x] Extract the clipboard logic into a separate `clipboard` module
- [x] Extract the application information extraction logic into a separate `app_info` module
- [x] Extract the macOS accessibility-related methods in a separate `accessibility` module
- [x] Extract the IPC logic into a separate `ipc` module
- [x] Extract the configuration parsing logic into a separate `config` module
- [X] Implement an alternative Linux subsystem based on the EVDEV and UINPUT, which is a necessary building block to port espanso to Wayland #287

### Deployment improvements
- [x] Create a `nightly` channel to rapidly ship bug fixes and improvements for testing, outside of the major release cycles.
- [x] Set up [clippy](https://github.com/rust-lang/rust-clippy) in the CI pipeline to fail when a warning is detected
[...]

PR #616, opened 2021-03-09
Update cargo.lock to allow building on homebrew

n8henrie:
[See](See): https://github.com/federico-terzi/espanso/issues/615


federico-terzi:
Hey @n8henrie,

Thank you for the help! I can't merge this one just yet, as I want to keep the 0.7.3 version fixed until #594 is ready. Given that I'm updating all dependencies in the process, there's a good chance this problem will be fixed without further action.
I'll keep the PR open until we can figure that out

Cheers :)

n8henrie:
Sounds good, thanks for the update.

PR #698, opened 2021-06-22
Document interaction of CLI & package script

katrinleinweber:
I realize that this might create problems, if untrusted code gets executed, but should the technical possibility of adding scripts to packages be documented?

For example in this manner?

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -42,3 +42,14 @@ espanso match exec <trigger>
 where `<trigger>` is the match trigger.
 
 This is useful if, for example, you want to trigger an expansion from a script.
+
+### Paths
+
+Besides the [overview command](configuration#structure) `espanso path`,
+you can also output the exact paths to specific components:
+
+```
+espanso path packages # or config, data or default
+```
+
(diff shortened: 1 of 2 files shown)


katrinleinweber:
Ah, no. The path doesn't get expanded after all, probably because it's not a shell but script type.

katrinleinweber:
I'm more confident in this PR after fd7f8f1. Before, 9ebb526 was definitely wrong, but the general concern about scripts in packages is still present.

federico-terzi:
Hey @katrinleinweber,

Thanks for your help! I thought about this issue extensively in the past, and my main concern was exactly what you highlighted: having scripts in public packages pose some significant concerns security-wise.
An option could be to allow scripts from public packages only if verified by the espanso team (as it currently is), but the process is not perfect yet.
After #594 is done (which should hopefully be around the end of summer), we will start to rewrite the package system from scratch with the insights we gained in version 1 and the feedback from the community (including this one!).

At that point, I'll be happy to continue this conversation so that we can improve the overall process together :)

Cheers :)

PR #1033, opened 2022-03-24
Fix broken "forms" link

DannyJJK:
The forms link led to a 404. I found the correct page and updated it


DannyJJK:
Closed this as I spotted there was already one open

PR #1043, opened 2022-03-27
chore: run gh action version update checks with dependabot

tobiasbueschel:
We could include this dependabot action to automatically check on a weekly basis whether there are any version updates available for the GitHub actions we are using.

This could also be configured for Cargo:
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: weekly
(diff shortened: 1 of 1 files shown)


tobiasbueschel:
Hi @federico-terzi also wanted to check on this PR - let me know what you think :) Thanks

AucaCoyan:
I arrived late to the party 😢 
dependabot is already working, so I have to close this PR. Thanks for the effort though!

PR #1062, opened 2022-04-06
Working "app-specific" for Wayland under Gnome

hseliger:
As I mentioned in a recent discussion (https://github.com/federico-terzi/espanso/issues/287#issuecomment-1085546978), using a Gnome extension it is possible to get the application filters working.

I created such a version and it is working perfectly for me (with respect to application filters).

2 limitations:

a) It currently only works under Gnome version 41. I have not used KDE in some 20 years (yep, I am *that* old) and don't know if xfce etc. use Wayland, so it would need someone else to make those ports. I also do not have different versions of Gnome (such as 42, which I guess would be key going forward) available for testing. **It requires the `dbus-1` library to be installed on the system.**

b) Since v. 41, Gnome enforces some "security" measures, which mean than applications cannot easily get details on other windows. So it requires a Gnome shell extension. I the discussion I mentioned "Window calls". That one only returns a JSON object, so espanso would have to take that apart. I created a version using rapidJSON, but it kind of bloats espanso and adds a non-standard library. So I opted to pimp (a.k.a. expand) the extension. I got it published on Gnome Extensions yesterday, so you will have to install "Window Calls Extended" (https://extensions.gnome.org/extension/4974/window-calls-extended/) to make application filters work in espanso.

Error handling still is quite simplistic. I also hope that I grasped the cargo build system sufficiently to get the Wayland version compiled without breaking anything else. There is a DEBUG switch in the code: 0 takes all debug stuff out, 1 writes many comments, and 2 even permits native.cpp to the compiled as program, so one can test it (`g++ -o dbus-test native.cpp -std=c++0x $(pkg-config dbus-1 --cflags) -ldbus-1 -Werror -Wall -Wextra`).

Have a look and merge if you feel it's safe.

– Hendrik


federico-terzi:
Hey @hseliger,

Thank you for the help! Impressive work. Please bear with me for delays in the review, I'm going through the busiest period of my life and I only have limited capacity for OS at the moment. I'll get back to you as soon as I can, I hope you'll understand :)

Cheers :)

lissy93 (approved):

hseliger:
Hi!
Just wondering if you're going to pull this change into the main branch. I'm still updating my fork with the changes for this mater, so far it went automatically, so I can use my fork under Wayland. Would just be easier – and maybe pull in contributors for other Wayland setups that Gnome/Fedora36 – if we had the initial Wayland support in the master branch.
Cheers,
Hendrik

lissy93 (approved):

waot:
Would this work for all Wayland? Let's say with Sway?
I don't have Gnome as a DE, so I was wondering if this would help my app-specific needs, as currently comes up empty three times with ''.

hseliger:
Since I don't use sway, I can only speak for my setup. Works well for me for many months now, but I use Gnome/Wayland/mutter. For other setups, you could only try and maybe help espanso by adapting code to work on your setup.

AucaCoyan:
Hi! this look good, can you switch the target branch from `master` to `dev`? 
thanks!
PD: [a nice gif](https://stackoverflow.com/questions/24159036/how-to-modify-a-pull-request-on-github-to-change-target-branch-to-merge-into) if you don't know how

federico-terzi:
Hi everyone! Finally got the chance to give this a proper review/thought, sorry for the late reply

First of all, thank you for the contribution, the work is impressive, and shows us that app-specific detection is possible in Wayland under Gnome

I don't think we should merge the PR as is, as the supported case is quite narrow, and brings additional complexity and dependencies to the project that would only benefit a small portion of users.

Thinking about the fragmented nature of Wayland (eg. Gnome, KDE, Sway, etc), and how each of those Desktop environments will require specific implementations to solve this problem, I think the long-term solution would be to introduce a (simple) plugin mechanism to add this functionality. This way, users could install Espanso + the specific plugin for their desktop environment.

This could be implemented in several ways. The easiest would be to have "App-Info" plugins be standalone executables that, when invoked, return the current app information as JSON to STDOUT:

For example:
```
{"title": "Google", "exec": "/usr/bin/firefox", "class": "Firefox"}
```

Espanso could then invoke this binary every time the app information is needed and parse the result. If the binary startup time is significant, we could switch to a different model where instead of invoking the binary on each call, Espanso would invoke it once and receive a streaming update of app information over time (eg. every 2/3 seconds):

```
{"title": "Google", "exec": "/usr/bin/firefox", "class": "Firefox"}
{"title": "Google", "exec": "/usr/bin/firefox", "class": "Firefox"}
{"title": "My document", "exec": "/usr/bin/libreoffice", "class": "LibreOffice Writer"}
```

This PR would serve as the perfect starting point to introduce the first Gnome-specific plugin

Thoughts? :)

hseliger:
I think this is a good idea. And you are fully right, the use case for this Gnome/Wayland solution is too narrow and a general plugin system should be more versatile. Wayland seems to require an extension to get to the data, but providing a command line utility to tap into that extension and feed the updates to expanso (or whoever may want them) should be quick to establish. And then it would indeed be easy for others to add KDE and others as well.
If I find the time, I'd happily have a go at the command line utility, although I'd probably go for C++ rather than Rust, just given experience.

PR #1277, opened 2022-07-02
feat: implement alt-code emulation. #988

federico-terzi:
Because the mechanism we use to detect keystrokes disables Windows' native Alt-code functionality, we try to bring it back by emulating it at the software level

https://github.com/espanso/espanso/issues/988


kpym:
Is this code included in the latest version? Because we see people still complaining (#1556, #1588).

appel:
This had me stumped for a bit as well until I figured out it was due to Espanso. Would it be an option to include `emulate_alt_codes: true` by default on new installs?

PR #1333, opened 2022-08-12
Clarify "Found duplicate 'name'" error

katrinleinweber:
After migrating from v0.7.3 using the compatibility mode, I had this folder structure (for some not relevant reason):

```
~/Library/Preferences/espanso $ exa --tree
├── config
│  └── default.yml # <= duplicate name
├── match
│  ├── packages
│  └── base.yml
├── user
└── default.yml    # <=duplicate name
```

I _think_ this resulted in the below error message. Is that so?

In any case, the message left me guessing which `name` is duplicated, and _where_ it is:

> [ERROR] unable to load legacy config
> 
> Caused by:
>     Found duplicate 'name' in '~/Library/Preferences/espanso/match/packages/…unrelated…/_manifest.yml', please use different names

Since that unrelated package contained quite a number of triggers with local variables, I thought at first that maybe _those names_ can't be duplicated, and renamed them. However, checking the `config.rs` code made me suspect the filenames as the problem source.

Is that so, and would in that case this rephrasing help?

### DIFF
--- a/espanso-config/src/legacy/config.rs
+++ b/espanso-config/src/legacy/config.rs
@@ -760,7 +760,7 @@ impl fmt::Display for ConfigLoadError {
             ConfigLoadError::InvalidYAML(path, e) => write!(f, "Error parsing YAML file '{}', invalid syntax: {}", path.to_str().unwrap_or_default(), e),
             ConfigLoadError::InvalidConfigDirectory =>  write!(f, "Invalid config directory"),
             ConfigLoadError::InvalidParameter(path) =>  write!(f, "Invalid parameter in '{}', use of reserved parameters in used defined configs is not permitted", path.to_str().unwrap_or_default()),
-            ConfigLoadError::NameDuplicate(path) =>  write!(f, "Found duplicate 'name' in '{}', please use different names", path.to_str().unwrap_or_default()),
+            ConfigLoadError::NameDuplicate(path) =>  write!(f, "Found duplicate filename, please use different names for your config files."),
             ConfigLoadError::UnableToCreateDefaultConfig =>  write!(f, "Could not generate default config file"),
         }
   }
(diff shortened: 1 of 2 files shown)


espanso-config/src/legacy/config.rs
```diff
@@ -760,7 +760,7 @@ impl fmt::Display for ConfigLoadError {
             ConfigLoadError::InvalidYAML(path, e) => write!(f, "Error parsing YAML file '{}', invalid syntax: {}", path.to_str().unwrap_or_default(), e),
             ConfigLoadError::InvalidConfigDirectory =>  write!(f, "Invalid config directory"),
             ConfigLoadError::InvalidParameter(path) =>  write!(f, "Invalid parameter in '{}', use of reserved parameters in used defined configs is not permitted", path.to_str().unwrap_or_default()),
-            ConfigLoadError::NameDuplicate(path) =>  write!(f, "Found duplicate 'name' in '{}', please use different names", path.to_str().unwrap_or_default()),
+            ConfigLoadError::NameDuplicate(path) =>  write!(f, "Found duplicate filename, please use different names for your config files."),
```
katrinleinweber, line 763: The `path` was not helpful in the resulting message. Upon deleting the package, the same error with a different package path appeared.

Thus, can either the _file_ paths of both duplicates be listed here, maybe? I'd be happy to let someone take over this PR, since I'm not very experienced in Rust.

federico-terzi:
Hey @katrinleinweber,

Thanks for the help! Interesting, indeed there is a relation between filename and name (if I recall correctly, the legacy espanso used the filename as "implicit" name if not explicitly specified), which might be confusing

That said, I suspect the problem is caused by the Espanso legacy/compatiblity mode not being compatible with the new package format (and there is no logic to correctly handle that, ouch)

So I don't think we should change "name" for filename here (as the filename is just a fallback), but we should definitely try to handle the new package format when loading the new Espanso packages (we can simply ignore the new packages when using compatibility mode). Would you like to give it a try or do you prefer me to take a look at it? :)
Hint: this is how we detect if a package is legacy: https://github.com/espanso/espanso/blob/feat/code-sign-on-macos/espanso-package/src/archive/default.rs#L104-L109

Cheers :)

katrinleinweber:
OK, thanks for explaining :-) I'll close this and open an issue, as I'm not equipped to try writing such a code patch myself.

Issue #1488, opened 2022-12-26
Autohotkey


AhmedThahir:
**Describe the bug**
Espanso does not expand when I use my custom keyboard layout script, but works when I disable my script.

**To Reproduce**
1. Use [Colemak autohotkey](https://github.com/omigo/colemak/blob/master/qwerty-to-colemak.ahk) script to use another keyboard layout.
2. Use espanso as usual
3. Typing `:date` using my keyboard layout does not expand to the current date

**Expected behavior**
I expect espanso to work the same as if I do not use my script

**Setup information**
- OS: Windows 10
- Version: 2.1.8

ernstki:
@AhmedThahir Federico will undoubtedly have deeper insights here, but try `:sagf` (`:date`, but using the keys from the Colemak layout) and see if you get a different result.

My gut feeling is that it would be very hard indeed for Federico to support your use case, especially if AutoHotkey and Espanso are working with the keyboard input buffer different levels. The AutoHotkey script is essentially waiting for certain key codes from the hardware and then inserting different characters into the input buffer. Espanso is probably receiving those same key codes, that is, the ones from the underlying QWERTY layout.

The result might be different if you were to use the "official" Colemak layout for Windows. Although I can't vouch for it, the Wikipedia article for Colemak [mentions](https://en.wikipedia.org/wiki/Colemak#cite_note-2) that there's one available for download from colemak.com. You'll need to decide for yourself whether installing third-party software that is literally monitoring every keystroke is something you're comfortable with. `;)`

jtroo:
Espanso ignores all events sent as a result of software using `SendInput`, e.g. when using:

- AutoHotkey
- PowerToys Keyboard Manager
- KMonad
- Kanata (my project)

The code responsible for that is here:

https://github.com/espanso/espanso/blob/4caf7f644267aa88b30939b3ced7fd94a8b92672/espanso-detect/src/win32/mod.rs#L188

There is a configuration option to stop ignoring these events, but then espanso just doesn't work correctly, from what I've found:

```yaml
# The default is true
win32_exclude_orphan_events: false
```

I've tried `backend: Clipboard` and `backend: Inject` and neither work correctly for me - the text expansion appears but then gets erased afterwards, with some extra leftover characters. E.g. when I type `gst` which expands to `git status`, I get see on screen:

- `gst`
- `git status`
- `ggst` - this is the final state

[...]

PR #1531, opened 2023-02-21
fix: make espanso work with other SendInput-using software

jtroo:
~~This is a proof of concept. It is quite unclean and in particular has a 100ms sleep which is certainly undesirable. Since I am unfamiliar with the design of espanso, I just hacked something together to show that it's possible to get this working. A better design should be used for the actual fix.~~

This is for #1488

~~Feel free to close whenever the issue is fixed by a better design.~~

The code has been updated to be reasonably responsive and is cleaned up from before. espanso-info may not be the right place to put the atomics though.


jtroo:
The implementation should be a lot nicer now, though code may be in the wrong place. E.g. the atomics code being in espanso-info.

federico-terzi:
Hey @jtroo, very interesting approach!
My main concern is that this solution might not be robust enough in the general case (including the `send_string`), but I'm glad it works well for you :)
For context, that software event filtering was added to prevent espanso from interfering with itself, potentially also falling into an infinite loop, eg:

```
- trigger: "hello"
  replace: "say hello"
```

The latter "hello" would be picked up, causing an infinite loop

jtroo:
In my testing the loop does not happen (Windows 10+11). Raw input does not catch unicode input as sent by `send_string` or clipboard paste. Increasing the expansion count in `send_string` makes the number of ignored inputs too large, which breaks expansion until the count is depleted.

Regarding the loop, tested with both below and worked fine:

```
  - trigger: "hello"
    replace: "hello hello"
```

and

```
  - trigger: "hello"
    replace: "hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello hello"
```

This code only really matters with the setting `win32_exclude_orphan_events: false`, which if enabled in espanso today, already completely breaks expansion 😅. Anyway, feel free to take inspiration from this and do a different fix whenever you have time.

melissis:
I tried your Windows builds, @jtroo, and unfortunately it didn't work.

I did follow the instructions on the release page.

So, to clarify, my issue is that espanso gets stuck and behaves oddly in certain programs (maybe only VSCode?) when certain AHK scripts are running in the background.
I noticed that, in VSCode, espanso is also very slow, with or without AutoHotKey running. Like, you can really see it deleting the characters. I don't know if there's a connection. I tried changing the Clipboard threshold and all the different backends, but it still doesn't work.

In writing this issue... I discovered that... it was this piece of AutoHotKey code that makes espanso not work in VSCode:
```
*^v::
*!v::
*#v::Send {Blind}.
```
Specifically, it's the `^v::` (Ctrl+V) part that causes issues.

jtroo:
> I tried your Windows builds, @jtroo, and unfortunately it didn't solve the issue for me.

Not too surprising that these changes didn't fix that issue! The fix in my code is for a specific use case of AHK (and other SendInput-using software). That use case being using AHK to remap the normal typing keys to other normal typing keys, e.g. to define a custom keyboard layout.

melissis:
> Not too surprising that these changes didn't fix that issue! The fix in my code is for a specific use case of AHK (and other SendInput-using software).

@federico-terzi Should I make a new issue?

federico-terzi:
I really appreciate your contribution, but I don't think we should move forward with this. The solution looks too patchy to be robust in the general case, and tangles dependencies quite a bit.
If a user wants to enable software-generated input, they should be able to turn on the `win32_exclude_orphan_events` config. It will pose risks of infinite loops though

I hope you'll understand :)

jtroo:
@federico-terzi I don't mind this being closed since I've been running my forked version for many months now, on a daily basis, and it works exactly as I need for my use cases. So I personally don't have any need for this to be merged.

Just do be aware that setting `win32_exclude_orphan_events: false`, at least back when I tested this in Feb, completely breaks espanso's functionality on Windows - it no longer works correctly **at all**. Maybe some other code has been added to fix #1488 since I last tried though? If not then I'm sure other users of `SendInput`-using software would appreciate some other fix that fits better into the project design, since it might be hard to discover the GitHub issue and my forked solution.

jmlucjav:
for the record, I have also been running jtroo's forked version, all working fine for me.

Geobert:
> Just do be aware that setting `win32_exclude_orphan_events: false`, at least back when I tested this in Feb, completely breaks espanso's functionality on Windows - it no longer works correctly **at all**.

I confirm this :-/

PR #1583, opened 2023-04-11
Allow overriding config dir path from env var

aslowwriter:
Heya folks. 
For my own setup I'd really like to be able to set the config path for espanso from an environment variable. 

I didn't really see a good place to add tests without having to basically build the test setup from scratch so I didn't add them, but if you want to see them added, point me where and I'll happily add them. 

Thanks for all the hard work :heart: 

Cheers :wave:

### DIFF
--- a/espanso-path/src/lib.rs
+++ b/espanso-path/src/lib.rs
@@ -17,8 +17,9 @@
  * along with espanso.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-use log::{debug, info};
+use log::{debug, info, warn};
 use std::{
+  env,
   fs::create_dir_all,
   path::{Path, PathBuf},
 };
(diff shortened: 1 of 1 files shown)


rafo:
That's precisely what I need to keep my dotfiles in one place `~/.config` on MacOs . Having a ESPANSO_CONFIG_DIR variable would be a dream.

Another solution could be to allow the Linux functionality `$XDG_CONFIG_HOME` under MacOS. 
Using https://espanso.org/docs/matches/organizing-matches/#imports is not really a substitute, because firstly the settings are missing and Espanso does not automatically notice when the match files are changed.

@savente93 with your solution, does espanso notice when something changes in the files?

januz:
@rafo Hey, I just installed `espanso` for the first time and was surprised that it doesn't honor `$XDG_CONFIG_HOME` for Mac OS. I would like to avoid having to use the standard path in the `Library/` dir.

Would you mind expanding on the solution you mention above. From looking at the documentation, it seems as though this would work for importing one or several `match` yaml files from anywhere (in our case `$XDG_CONFIG_HOME/espanso`). But does this also work for the `config`?

Thanks!

rafo:
@januz Sure!

1. On macOS Sonoma open:
```/Users/YOURUSERNAME/Library/Application Support/espanso/config```
2. Copy your ```config``` folder to the espanso path under ```.config/espanso/```
3. Add a file with the name ```default.yml``` and insert:
```
# Import other matches from external locations
imports:
  - "/Users/YOURUSERNAME/.config/espanso/"
```
4. Restart espanso.
5. Open y new terminal window.
6. Test the config path with ```espanso edit```
7. Enjoy a *sane* config folder for your dot files.

```
   O
  o
:°
```

PR #1697, opened 2023-09-07
Update package dependencies for Debian 12+ and Ubuntu 22.10+

### LINKED ISSUE #1662: package install fails on debian trixie
**Logs**

```
The following packages have unmet dependencies:
 espanso-wayland : Depends: libwxgtk3.0-gtk3-0v5 but it is not installable
                   Depends: libwxgtk3.0-gtk3-0v5 (>= 3.0.4+dfsg) but it is not installable
                   Depends: libwxbase3.0-0v5 (>= 3.0.4+dfsg) but it is not installable
E: Unable to correct problems, you have held broken packages.
```

**Setup information**
- OS: What OS are you using? debian trixie (testing)
- Version: which version of Espanso are you running? attempting to install 2.1.8

**Analysis**

the issue is that debian does not have these versions of wx any longer:

```
$ apt search '^libwx(base|gtk3)'
Sorting... Done
Full Text Search... Done
libwxbase3.2-1/testing,unstable,now 3.2.2+dfsg-2 amd64 [installed,automatic]
  wxBase library (runtime) - non-GUI support classes of wxWidgets toolkit

libwxgtk3.2-1/testing,unstable,now 3.2.2+dfsg-2 amd64 [installed,automatic]
  wxWidgets Cross-platform C++ GUI toolkit (GTK 3 runtime)

libwxgtk3.2-dev/testing,unstable,now 3.2.2+dfsg-2 amd64 [installed,automatic]
  wxWidgets Cross-platform C++ GUI toolkit (GTK 3 development)
```

however, this seems to be more than simply a packaging issue. when i extract the deb and manually execute the binary, it fails because it is linked to `libwx_gtk3u_html-3.0.so` which doesn't exist on my system. i'm not sure if there were breaking APi changes between these versions of libgtk.

### LINKED ISSUE #1674: Libraries libwxbase and libwxgtk3 are deprecated in Debian 12 - unable to install the provided .deb file
When trying to install the .deb packages on Debian 12 (stable) I get the error:

The following packages have unmet dependencies:
 espanso-wayland : Depends: libwxgtk3.0-gtk3-0v5 but it is not installable
                   Depends: libwxgtk3.0-gtk3-0v5 (>= 3.0.4+dfsg) but it is not installable
                   Depends: libwxbase3.0-0v5 (>= 3.0.4+dfsg) but it is not installable
E: Unable to correct problems, you have held broken packages.

There are newer version of those libraries installed in the system. It seems that dependencies in the package should be updated.

rcalixte:
* Fixes #1674 
* Fixes #1662

### DIFF
--- a/Compilation.md
+++ b/Compilation.md
@@ -72,19 +72,22 @@ This will create the `Espanso.app` bundle in the `target/mac` directory.
 
 # Linux
 
-Espanso on Linux comes in two different flavors: one for X11 and one for Wayland. 
+Espanso on Linux comes in two different flavors: one for X11 and one for Wayland.
 If you don't know which one to choose, follow these steps to determine which one you are running: https://unix.stackexchange.com/a/325972
 
 ## Compiling for X11
 
 ### Necessary packages
 
-If compiling on Ubuntu X11:
+If compiling on a version of Ubuntu X11 before 22.04 (including 22.04):
 * `sudo apt install libx11-dev libxtst-dev libxkbcommon-dev libdbus-1-dev libwxgtk3.0-gtk3-dev`
+
+If compiling on a version of Ubuntu X11 after 22.04:
+* `sudo apt install libx11-dev libxtst-dev libxkbcommon-dev libdbus-1-dev libwxgtk3.2-dev`
 
 ### AppImage
 
-The AppImage is a convenient format to distribute Linux applications, as besides the binary, 
+The AppImage is a convenient format to distribute Linux applications, as besides the binary,
 it also bundles all the required libraries.
 
 You can create the AppImage by running (this will work on X11 systems):
--- a/espanso/Cargo.toml
+++ b/espanso/Cargo.toml
@@ -27,20 +27,20 @@ wayland = ["espanso-detect/wayland", "espanso-inject/wayland", "espanso-clipboar
 modulo = ["espanso-modulo", "espanso-clipboard/avoid-gdi", "espanso-ui/avoid-gdi"]
 
 [dependencies]
-espanso-detect = { path = "../espanso-detect" } 
-espanso-ui = { path = "../espanso-ui" } 
-espanso-inject = { path = "../espanso-inject" } 
-espanso-config = { path = "../espanso-config" } 
-espanso-match = { path = "../espanso-match" } 
-espanso-clipboard = { path = "../espanso-clipboard" } 
-espanso-info = { path = "../espanso-info" } 
-espanso-render = { path = "../espanso-render" } 
-espanso-path = { path = "../espanso-path" } 
-espanso-ipc = { path = "../espanso-ipc" } 
+espanso-detect = { path = "../espanso-detect" }
+espanso-ui = { path = "../espanso-ui" }
+espanso-inject = { path = "../espanso-inject" }
+espanso-config = { path = "../espanso-config" }
+espanso-match = { path = "../espanso-match" }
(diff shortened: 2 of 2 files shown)


rcalixte:
> Hey @rcalixte! Thanks for the help :) (and sorry for the late reply)

No need for apologies!


> The PR looks good, but we also need to update the CI scripts/dockerfile to include the updated packages: https://github.com/espanso/espanso/blob/dev/.github/scripts/ubuntu/Dockerfile
> 
> I'm wondering if there's a way we can create a deb package that is compatible with ubuntu 22+ _and_ <22, any ideas?

This is using 18.04 as the base image and therefore will only have 18.04 repositories available. We would need to create a new Dockerfile for either 22.10+ (or wait until the next LTS in a few months). But the resulting .deb file _should_ be compatible with both versions since the package name changed but the underlying library is mostly the same? The local distribution would be able to handle package dependency resolution using the repositories that are available.


Is there a way to test a build against this pull request?

rcalixte:
@federico-terzi If you can fix the build issues in the source repository, I can rebase the forked repository with the changes and test a build with these changes. I tried running the build and there are issues with some of the Rust components but once they are resolved, we can reliably test the results before merging. Let me know if this sounds like a good plan!

federico-terzi:
@rcalixte I'll try to do it in the upcoming week :)

federico-terzi:
> @federico-terzi If you can fix the build issues in the source repository, I can rebase the forked repository with the changes and test a build with these changes. I tried running the build and there are issues with some of the Rust components but once they are resolved, we can reliably test the results before merging. Let me know if this sounds like a good plan!

Hey @rcalixte :) Build errors on the `dev` branches should be fixed, so feel free to proceed whenever you have the chance :)

Thanks for your help!

rcalixte:
@federico-terzi I was able to get it working. I tested both x11 and Wayland packages on Ubuntu 22.04 and Debian 12. The fix ended up being to remove the `$auto` keyword from the Cargo.toml file (for now?). The `$auto` keyword made cargo-deb add hard-coded dependencies that would only be for one or the other. So I pulled out the resulting dependencies and removed that keyword.

The release page with the deb build assets is [here](https://github.com/rcalixte/espanso/releases/tag/v2.2.0) until the fork is deleted at some point. You can download and test with those.

Essentially, this ended up being a tradeoff for simplicity. The other option is to rework the entire build process to support both package variants until the older ones are end of life. I would lean toward waiting for the end of life for the older distributions and then updating the Cargo.toml to re-add the `$auto` keyword once it would be safe to do so. This means only having to do the change to the build process once and in the future with the advantage of proper planning. Provided that the dependencies don't change too much, this isn't a bad option.

AucaCoyan (approved):
Looks great to me! Great job!
Thank you 🙏🏼

federico-terzi:
Thanks! Seems to work correctly on Ubuntu 20 as well, good to go

ernstki:
> Thanks! Seems to work correctly on Ubuntu 20 as well, good to go

It appears that Ubuntu 22.04 (and derivatives) will need to [download 2.1.8](https://github.com/espanso/espanso/releases/tag/v2.1.8) from the releases page and use that instead (or compile from source), due to the unsatisfiable dependency on `libwxgtk` and `libwxbase` version 3.2, since the Jammy repositories [only have 3.0](https://packages.ubuntu.com/search?keywords=libwxbase&searchon=names&suite=jammy&section=all).

PR #1704, opened 2023-09-13
Update SECURITY.md

brettporter:
Updated links for version 2, based on my understanding of the code. Would this be correct and sufficient to address the TODO at the top of the file, or are there other areas that have changed that you had in mind?

Happy to help update the document if I can!

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -26,12 +26,13 @@ due to the *global key detector* we mentioned before. The good news is, **it's n
 
 While espanso detects key presses as a keylogger would do,
 **it doesn't log anything**. Moreover, to further reduce risks, espanso only
-stores in memory the last 3 chars by default (you can change this amount by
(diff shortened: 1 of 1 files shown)


jaroslaw-weber (approved):

AucaCoyan (approved):

federico-terzi:
Thanks! Unfortunately, this information is not fully accurate in version 2. As far as I remember (it's been a long while since I touched it!), the main differences are:
* Using regex matches causes the system to store up to 30 chars by default: https://github.com/espanso/espanso/blob/dev/espanso-match/src/regex/mod.rs#L55
* The regular matches buffer are controlled by [this value ](https://github.com/espanso/espanso/blob/72b516c0ef53ab10dc5289da3355f300af875cda/espanso/src/cli/worker/config.rs#L174), which by default is set to 5 ([source](https://github.com/espanso/espanso/blob/72b516c0ef53ab10dc5289da3355f300af875cda/espanso-config/src/config/resolve.rs#L249))

If someone could update this PR to reflect this new information, it would be amazing! :)

PR #1782, opened 2023-12-13
:memo: Update the path for windows portable compilation

AucaCoyan:
Just add the path for portable windows, that was missing on `Compilation.md`


federico-terzi:
Thanks!

PR #1786, opened 2023-12-15
:green_heart: Add `fail-fast: true` to stop the CI with fails

AucaCoyan:
Just a tiny line change, 
```yaml
    fail-fast:true
```
Does that when any check is red, stops the workflow entirely, so don't waste resources that we don't need 😄

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ env:
 jobs:
   build:
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
 


federico-terzi:
Hey @AucaCoyan , thanks for your contribution!
I'm a bit torn on this one, as sometimes it's quite helpful to have all targets continue their checks even if one of them fails. For example, there are some errors that are platform-specific: in those cases, it's very handy to see the result across all targets to fix the problems of all platforms at once. If we enable this, fixing problems across multiple platforms would require a slower sequential process (eg. fix the problems on Linux, see if it builds, fix the problems on Windows, see if it builds, etc)

Thoughts?

AucaCoyan:
Mmmm yes, you are right. As I say in spanish, "I have a library in favour and a library against it" lol
Both sides are correct. Let's keep it classic until we need to! 💪🏼

PR #1694, opened 2023-09-03
Bump dependencies

eclipseo:
I only touched those I needed to build on Fedora.

-> enum-as-inner = "0.6.0"
-> mockall = "0.11.4"
-> smithay-client-toolkit = "0.16.0" (version 0.17.0 has too many API changes due to wayland-rs 0.30 rewrite)
-> html2text = "0.6.0"
-> path-slash = "0.2.1"
-> include_dir = { version = "0.7.3", features = ["glob"] }
-> zip = "0.6.6"
-> simplelog = "0.12.1"
-> dialoguer = "0.10.4"
-> notify = "4.0.17" replaced by notify-debouncer-mini = "0.4.1"

PR #1796, opened 2023-12-19
:arrow_up: Upgrade `enum-as-inner` across crates

AucaCoyan:
WIP

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      # - uses: Swatinem/rust-cache@v1
+      - uses: Swatinem/rust-cache@v1
       - name: Check formatting
         run: |
           rustup component add rustfmt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: Swatinem/rust-cache@v1
+      # - uses: Swatinem/rust-cache@v1
       - name: Check formatting
         run: |
           rustup component add rustfmt
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ DerivedData
 venv/
 
 .vscode/
+*.profraw
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -289,7 +289,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "29c36c619c422113552db4eb28cddba8faa757e33f758cc3415bd2885977b591"
 dependencies = [
  "proc-macro2",
- "quote 1.0.9",
+ "quote 1.0.33",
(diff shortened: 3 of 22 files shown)


AucaCoyan:
Every other build is ready, but for some reason fails in windows. I don't understand the error 😢

federico-terzi:
@AucaCoyan Thanks! I'm wondering if the windows error is caused by the Rust modules caching in the CI... Try to comment out this line: https://github.com/espanso/espanso/blob/dev/.github/workflows/ci.yml#L30

If it passes, we might need to update the Rust cache version :) https://github.com/Swatinem/rust-cache

AucaCoyan:
Sadly it wasn't that,
Looking the files, i found that there are 2 dependencies that uses `windows` crate, one  "windows 0.51.1", and the other  "windows 0.24.0"
The old one is needed by `winrt-notification`, so I'll investigate if I can upgrade the windows crate to the same version

AucaCoyan:
Aaaand nope, that is not it. Though it is still a better/newer dependency, so we can keep it

AucaCoyan:
I rolled back the changes on the `espanso` crate and it worked. I see that we need to make this step first, and then in the forseable future update `enum-as-inner` in the future.
There is another PR already with this and another dependencies #1694 

If this PR gets landed after a review @federico-terzi , I would continue there!

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: Swatinem/rust-cache@v1
+      # - uses: Swatinem/rust-cache@v1
```
federico-terzi: Disabling caching on the CI would bloat the times significantly (even up to 30 min at a time), can we revert it? :)

.gitignore
```diff
@@ -37,4 +37,5 @@ DerivedData
 
 venv/
 
-.vscode/
\ No newline at end of file
+.vscode/
+*.profraw
```
federico-terzi, line 41: Why is this necessary? :)

federico-terzi (commented):
Thanks! Left a couple of comments :)

.gitignore
AucaCoyan: Because `.vscode/` is the folder where the Visual Studio Code saves the files of the editor. I think we shouldn't commit personal config files to the repo. Note that this prevents adding the folder .vscode when you `git add .`, but we can add a file with `--force` if needed (for example recommending and extension, or doing some build task, and so on)

`*.profraw` files are the result of doing coverage tests. [Elkiwa left a great guide on discord](https://discord.com/channels/884163483409731584/1013902234469269554/1186670417360977921)

AucaCoyan:
Well, github actions failed because it didn't found the github server
```
Error: nodename nor servname provided, or not known (api.github.com:443)
```
Trying with another empty commit

AucaCoyan:
Done! I also upgraded the rust edition from 2018 to 2021 😄

federico-terzi:
Thanks! Looks great :)

PR #1797, opened 2023-12-20
:arrow_up: Upgrade `cargo-make` to version `0.37.5`

AucaCoyan:
I updated `cargo-make` for the latest version.

```
cargo-make --version 0.34.0
```

- [x] updated the docs


AucaCoyan:
Well, another branch with more C++ problems 😢

AucaCoyan:
I couldn't make it work. Can somebody with C++ experience try to read the error and see whats happening?
Thanks in advance!

federico-terzi:
> I couldn't make it work. Can somebody with C++ experience try to read the error and see whats happening?

Isn't it because you commented out a header from the C file? :)

AucaCoyan:
Whoops, I comment that line as a blind shot if it works, but had (obviously) no success. I revert back the comment but the CI is still getting an error.
Can you read it again when you have some time to invest? thanks! 😄

AucaCoyan:
Strange, I setup a Debian 12 VM in my machine and it builds and passes the tests correctly.

Maybe github is failing to build randomly?

AucaCoyan:
Done!
Turns out that 
```
cargo make build-binary --env NO_X11=true
```
is not the same as
```
cargo make --env NO_X11=true build-binary
```
because `cargo make` forwards the rest of the arguments when you name a command, so in the first sample `--env NO_X11=true` was given to `cargo`, not to the build script.

I also logged the default features, so when you are in Wayland and you see a log telling you
```
Using X11 default feature
```
it means it is not receiving the argument correctly

This PR is ready to merge @federico-terzi !

federico-terzi:
Thanks! Great catch

Issue #1760, opened 2023-11-29
Bad encoding on MacOS for markdown


devcrafting:
**Describe the bug**
On MacOS, when I use a trigger with markdown replacement with accentuated characters (like "é"), it is not rendered correctly.

**To Reproduce**
1. Add a trigger with markdown "éléphant" for exemple in config
2. Use that trigger on MacOS

**Expected behavior**
It should write "éléphant" and not weird encoding instead of "é" characters

**Screenshots**
Not necessary

**Logs**
The replacement work, nothing special in logs.

**Setup information**
- OS: MacOS (tested both on M1 and M2, same result)
- Version: 2.1.8

devcrafting:
wondering if mardown crate 0.3.0 seems to have lots of problems, perhaps possible to upgrade to 1.0.0-alpha version could help ? I will try to test quickly (but I do not have a MacOs for dev)

devcrafting:
I tried on a team mate MacOS, upgrading mardown crate to 1.0.0-alpha version does not solve the issue :(

Also tried with the html replacer, it is the same problem, so I will investigate in this direction (since mardown is converted to HTML with mardown crate, sounds logic there is something here)

devcrafting:
We manage to restrict the bug to div contenteditable in webkit-based browser (tested with Chrome, Safari, Brave). It works in textboxes, and even in div contenteditable in Firefox. So probably a problem with the backend ? (we tried Inject and Clipboard, does not change anything)

AucaCoyan:
Mmm this is where @federico-terzi can tell better what happening

kpym:
What do you mean by "markdown replacement"? Are you talking about a replacement that you activate while editing a markdown file?

devcrafting:
markdown replacement is when you this type of trigger :

```
- trigger: ":test"
  markdown: "éléphant"
```

This is not correctly transformed to "éléphant" on MacOS in webkit-based browsers. The "é" are badly encoded with ugly characters...

By the way, we discovered another problem on MacOS : the formatting is taken in account in webkit-based browsers, but not in Firefox :(...

So we did a quick experiment with a colleague : https://github.com/mmetesreau/espanso/tree/cocoa-clipboard-html and it seems to solve both problems on encoding in webkit and formatting in Firefox. Tested with Mac Intel version. Do not know exactly if this small change can have huge impacts on other usage ?

AucaCoyan:
[...]

PR #1807, opened 2024-01-05
feat(clipboard): add representation for type NSHTMLPboardType in NSPasteboard 

mmetesreau:
As mentioned in the discussion #1760 by @devcrafting, it looks like that any HTML div tag with the attribute `contenteditable` in webkit-based browsers on MacOs is not able to "consume" RTF text which leads to the fact that raw text is used instead.

A solution which seems to fix the issue is to add a third representation of the data to be pasted for the type NSHTMLPboardType (representation consumables by the browsers in the case specified above).

We have tested it on the latest MacOS version Sonoma (Intel and M2 processor) ensuring that RTF text is still used for applications like TextEdit.

### DIFF
--- a/espanso-clipboard/src/cocoa/native.mm
+++ b/espanso-clipboard/src/cocoa/native.mm
@@ -79,7 +79,8 @@ int32_t clipboard_set_html(char * html, char * fallback_text) {
                                   documentAttributes:nil];
 
   [pasteboard setData:rtf forType:NSRTFPboardType];
-
+  [pasteboard setString:nsHtml forType:NSHTMLPboardType];
+  
   if (fallback_text) {
     NSString *nsText = [NSString stringWithUTF8String:fallback_text];
     [pasteboard setString:nsText forType:NSPasteboardTypeString];


AucaCoyan:
Hi! Thank you very much for the PR. I'm all in for it!
@federico-terzi can you give it a glance? Thanks!

federico-terzi:
Thanks! Looks great

Waiting on the CI to pass before merging :)

AucaCoyan:
CI didn't like it 😢 , can you make the change please?

federico-terzi:
@mmetesreau Could you please pull and repush from dev? I've fixed a CI problem :)

mmetesreau:
@federico-terzi  rebased from dev.

federico-terzi:
Thanks!

Gargooie:
Can it be done for Chrome browser? It works for Safari and TextEdit but for any other broswer RTF text is being pasted as raw.

AucaCoyan:
That's nice, Chrome is the mayor browser. Would you mind creating an issue asking for this? 😄 
(don't worry about the discussions text)

PR #1831, opened 2024-01-29
:arrow_up: upgrade `fs_extra` and `html2text`

AucaCoyan:
So we don't have the warning that rust is going to deprecate the crates

(the 2 html5ever and xml5ever are dependencies of `html2text`)


federico-terzi:
Thanks!

PR #1847, opened 2024-02-12
:truck: Reorganize the repo in `crates` and `src` folders

AucaCoyan:
Hi, I ran a little experiment here, I separated the binary `espanso` into a `src` folder, and all other crates into `crates/` subfolder.
I belive this a little more organized.
Let's see if CI crashes... No! woah, it surprises me that nothing broke yet 🤣 

## Reason for this change

A change this big should have a little explanation on why:

points in favor:
- `src` actually corresponds to the only binary in the app, while `crates` is populated with all the libraries
- removes the "empty" crate that is workspace today. It's just a container of a lot of crates, but it doesn't have any other function than that
- we can have a nice `Cargo.toml` right up in the root of the repo
- we have less folders, in just 1 screen you can see all the files.
- I made the changes and CI didn't raised an eyebrow

points against it:
- it may have some little time to get used to
- I believe it will have many merge conflicts for all the current PRs (not sure about that, but it is some work if that is true)
- this architecture is less common, I see some big crates organized in the current folder structure (serde, tokio, clap)


AucaCoyan:
As a suggestion, @phorcys420 told me in discord, how about we remove the `espanso-` part of the crates inside the folder? it would turn from this:
image
into this:
image

so if you want to find a file you dont need the last `espanso` in
`espanso/crates/espanso-info/Cargo.toml` for example

What do you think?

AucaCoyan:
I'll close this PR for the time being, but in the future I'm interested in having it merged!

PR #1779, opened 2023-12-13
chore: :rotating_light: more `clippy` pedantic lints

AucaCoyan:
Hi! 
from #1775 I had a good chunk of fixes landed, so kept going!

used command
```console
cargo clippy --all-targets -- -W clippy::pedantic -A clippy::too_many_lines -A clippy::needless_pass_by_value -A clippy::missing_errors_doc -A clippy::missing_panics_doc -A clippy::cast_possible_wrap -A clippy::module_name_repetitions -A clippy::match_same_arms -A clippy::unnecessary_wraps -A clippy::cast_possible_truncation -A clippy::must_use_candidate -A clippy::similar_names -A clippy::struct_excessive_bools -A clippy::wildcard_imports -A clippy::items_after_statements -A clippy::used_underscore_binding -A clippy::borrow_as_ptr -A clippy::no_effect_underscore_binding -A clippy::cast_sign_loss -A clippy::ptr_as_ptr
```

or,  you can see the lints on VS Code with the config on `settings.json`:

<details>

```json
  "rust-analyzer.check.command": "clippy",
  "rust-analyzer.check.extraArgs": [
    // "--all-targets", // it doesnt work with this flag
    "--",
    "-W",
    "clippy::pedantic",
    "-A",
    "clippy::too_many_lines",
    "-A",
    "clippy::needless_pass_by_value",
    "-A",
    "clippy::missing_errors_doc",
    "-A",
    "clippy::missing_panics_doc",
    "-A",
    "clippy::cast_possible_wrap",
    "-A",
    "clippy::module_name_repetitions",
    "-A",
    "clippy::match_same_arms",
    "-A",
    "clippy::unnecessary_wraps",
    "-A",
    "clippy::cast_possible_truncation",
    "-A",
    "clippy::cast_sign_loss",
    "-A",
    "clippy::ptr_as_ptr",
    "-A",
    "clippy::as_conversions",
    "-A",
    "clippy::used_underscore_binding",
    "-A",
    "clippy::no_effect_underscore_binding",
    "-A",
    "clippy::borrow_as_ptr",
    "-A",
    "clippy::must_use_candidate",
    "-A",
    "clippy::similar_names",
    "-A",
    "clippy::struct_excessive_bools",
    "-A",
    "clippy::wildcard_imports",
    "-A",
    "clippy::items_after_statements",
    "-A",
    "clippy::used_underscore_binding",
    "-A",
    "clippy::module_name_repetitions"
  ],
```

</details>

PR #1775, opened 2023-12-11
Improve code quality opting-in some `clippy` pedantic lints

AucaCoyan:
Hi! I love clippy lints, and `cargo clippy -- -W clippy::pedantic` gives us a lot of output.
Not everything is useful, but there are some nice things here and there.

the running command I am using

```console
cargo clippy -- -W clippy::pedantic -A clippy::too_many_lines -A clippy::needless_pass_by_value -A clippy::missing_errors_doc -A clippy::missing_panics_doc -A clippy::cast_possible_wrap -A clippy::module_name_repetitions -A clippy::match_same_arms -A clippy::unnecessary_wraps -A clippy::cast_possible_truncation -A clippy::must_use_candidate -A clippy::similar_names -A clippy::struct_excessive_bools
```

PR #1853, opened 2024-02-16
:rotating_light: add `clippy` lints to `Cargo.toml`

AucaCoyan:
After #1775 and #1779 landed, I only needed to add the lints to clippy so we can check the quality of the code.
The problem was that I couldn't found a way at that time that with simply `cargo clippy` on a terminal rust gather all the lints I did for those PRs.
And recently I found that there [was a RFC](https://rust-lang.github.io/rfcs/3389-manifest-lint.html) to `Cargo.toml`

I tried that and worked since rust 1.75! So I'm implementing that in our codebase


AucaCoyan:
Ok, fixed all the lints of macos via CI 👍🏼 
ready for review!

AucaCoyan:
This was merged with the #1856 PR, so it's no longer needed

PR #1872, opened 2024-02-27
:package: update `Cargo.lock`

AucaCoyan:
Update the cargo lock


AucaCoyan:
turns out we shouldn't have a cargo.lock in a application, says [gitignore](https://github.com/github/gitignore/blob/main/Rust.gitignore)
so I'm removing that

AucaCoyan:
I will close this because it spread on many different problems, that can't be addressed with one PR.
I created an issue to keep track of the tasks #1953

n8henrie:
@AucaCoyan Just FYI -- the gitignore project seems to have that backwards. `Cargo.lock` should be checked in for applications (since what you want is a working binary at nearly all costs) but ignored for libraries (to allow flexibility with other dependencies, permit the automatic "latest version" behavior to pick up security updates, etc.) -- this is similar to lockfiles in other languages IME.

The URL they cite as a reference doesn't seem to have any relevant content: https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html , and it looks like github/gitignore hasn't been updated in 2 years. I would start an issue there, but they have issues disabled.

There has been some link rot, but relevant reading:
- https://github.com/rust-lang/cargo/issues/315
- [Why do binaries have Cargo.lock in version control, but not libraries?](https://web.archive.org/web/20171006215241/http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries)
- [Cargo.toml vs Cargo.lock](https://web.archive.org/web/20141109124342/http://doc.crates.io/guide.html#cargo.toml-vs-cargo.lock)

AucaCoyan:
> the gitignore project seems to have that backwards

Yes, exactly! I fell on that well and lead me to a dead-end street. A buddy of discord tapped me on the shoulder and point that out 😆 . Fun times.
Having said that, I created the issue because if you delete the Cargo lock and rebuild the app, it downloads the latest version of many dependencies and points to some weak issues in the code. One of them is related to `chrono` and the time security bug
Thank you for the very thorough revision of the PRs! I hope we keep seeing you often 😃

PR #1873, opened 2024-02-27
:construction_worker: make clippy run before the build

AucaCoyan:
I noticed that the CI does this things in order:
- checks out the repository
- uses `rust-cache`
- checks the formatting
- installs the linux dependencies
- installs `rust-script` and `cargo-make`
- caches `wxWidgets` builds
- builds the binary
- and then runs clippy

And thought that clippy should be runned just after the formatting, this PR fixes that


AucaCoyan:
I see. You need wxWidgets to compile _and then_ run `cargo clippy`

PR #1907, opened 2024-03-18
:fire: remove `Cargo.lock`

AucaCoyan:
Cargo lock is not actually needed in rust apps

more info here
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html


AucaCoyan:
I see why it was failing: the moment you delete `Cargo.lock` cargo rebuilds it differently, and throws the 
```
error: use of deprecated associated function `chrono::TimeDelta::seconds`: Use `TimeDelta::try_seconds` instead`
```
This happens because it uses the most modern dependencies available. Because of that, wxWidgets are for some reason incompilable. We need to search for the cause and fix it, but it's deep in a 3600 lines long file

morgan-dgk:
Hey @AucaCoyan, unless I'm reading it wrong the linked doc seems to suggest _we should check-in cargo.toml to git/VCS._:

> When in doubt, check Cargo.lock into the version control system (e.g. Git).

AucaCoyan:
You are right! I first read [this](https://github.com/espanso/espanso/pull/1872#issuecomment-1996259072) comment from the `.gitignore` startup, and turns out that comment is wrong AND I misread the rust docs 🤣 .
Thank you for checking!
I'm closing this and restoring #1872

PR #1917, opened 2024-03-28
:sparkles: add schemas

AucaCoyan:
I've been playing around with schemas in yaml, and we can have a schema for the configuration.
Not everywhere is useful, but there are some cases that it highlights the error:

### tasks
- [x] schema for `base.yml`
- [x] schema for `config.yml`

### how to use it:
In VS Code, if install the [extension Yaml by Red Hat](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)
and you put this line in your matches file:
```
# yaml-language-server: $schema=https://raw.githubusercontent.com/AucaCoyan/espanso/add-yaml-schemas/schemas/match.schema.json
```
and this in your config file
```
# yaml-language-server: $schema=https://raw.githubusercontent.com/AucaCoyan/espanso/add-yaml-schemas/schemas/config.schema.json
```

you should get errors like this:
image

image

and autocomplete like this:
image


AucaCoyan:
PR is ready to merge, but I'll wait for the reviewers to give me feedback about it 😄

espanso/src/cli/cmd.rs
```diff
@@ -49,7 +50,7 @@ fn cmd_main(args: CliModuleArgs) -> i32 {
     IPCEvent::ToggleRequest
   } else if cli_args.subcommand_matches("search").is_some() {
     IPCEvent::OpenSearchBar
-  } else if cli_args.subcommand_matches("search").is_some() {
+  } else if let Some(subcommand_matches) = cli_args.subcommand_matches("search") {
```
n8henrie, line 53: Why add an unused variable (and then disable the lint suggesting that you don't do so)? Shouldn't this be left as `is_some()` like the others?

AucaCoyan: Yeah, but `clippy` went mad if I left this as it is. We are checking if `cli_args` is `search` twice, and also we should take the path of `match`, which exhausts every possible option.
This section is more a work in progress, I think

smeech:
In VSCodium the (date extension) `locale:` item results in a `Property locale: is not allowed` which I assume is because it is absent from the match schema.

AucaCoyan:
Yup, that is probably correct. I'll fix that rightaway, since is an easy bandage 👍🏼

AucaCoyan:
it's done in #2040 . 
For some reason it's not working yet with the schema link `https://raw.githubusercontent.com/espanso/espanso/dev/schemas/match.schema.json` but it eventually refresh the cache and will

chrisgrieser:
very nice, thank you!

btw, if you add the file to the schema store, the schema gets applied automatically for the correct file, without the need to add the `schema` header to every file.
https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#how-to-add-a-json-schema-thats-self-hostedremoteexternal

AucaCoyan:
Hi Chris!
It would be cool to have it, yeah, but I think it would only apply to the `fileMatch` we specify (for example `default.yml` and `base.yml`) that's it.
It doesn't load the schema on other (custom) files _unless you copy and paste_ the `schema:` line. We shouldn't do a `*.yml` because a lot of yaml files coexist in a system.

If a PR is made, we welcome it and we would be happy to do anything needed, but because the default config already have the yaml schema link, it doesn't add anything new, just another layer of protection if the user accidentally deletes `# yaml-language-server: ...`. Again, this only works on the `default` and `base` files.

If anyone else want to take it, I can show directions!

PR #1918, opened 2024-03-29
Fix build with wx-config-qt

astehlik:
When installing the [wxwidgets-qt5](https://archlinux.org/packages/extra/x86_64/wxwidgets-qt5/) Package in Arch, the binary `wx-config` does not exist. It only installs `/usr/bin/wx-config-qt`.

To fix the build when this package is used it is added as a possible location when looking for the `wx-config` command.

### DIFF
--- a/espanso-modulo/build.rs
+++ b/espanso-modulo/build.rs
@@ -440,24 +440,24 @@ fn macos_link_search_path() -> Option<String> {
 #[cfg(target_os = "linux")]
 fn build_native() {
   // Make sure wxWidgets is installed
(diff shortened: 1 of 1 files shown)


AucaCoyan:
Hi! Thanks for submitting a PR. It seems the CI failed because of clippy. I did the fix in #1921 , can you update your branch and see if the CI turns green? Thanks!

astehlik:
Thank you for the quick response. I updated my branch to the latest dev :+1:

AucaCoyan:
Thank you! Merged and it built properly 🚀

Issue #1945, opened 2024-04-13
Missing return in `espanso-detect/src/mac/native.mm:detect_initialize` leads to runtime crash


n8henrie:
### Description

Hello,

PR incoming ... EDIT: https://github.com/espanso/espanso/pull/1946

I've been working on fixing espanso (when installed via nix on aarch64-darwin) for nearly a year now: https://github.com/NixOS/nixpkgs/issues/247162 (although I only opened the issue about 8 months ago).

I use rust but don't have a background in objective C or C++, so the issue has taken a *long* time for me to sort out, but I think I finally have it figured out:

the signature for the function in question suggests a `void *` return value but the function body lacks a return statement: https://github.com/espanso/espanso/blob/dev/espanso-detect/src/mac/native.mm

In both C++ and objective C this should result in the warning:

```
warning: non-void function does not return a value [-Wreturn-type]
```

For whatever reason it still successfully compiles and runs natively on aarch64-darwin (with `cargo build`) and doesn't even result in any compilation errors, but from what I can tell, because a `void *` function is not the same as a `void` function, this results in undefined behavior: https://stackoverflow.com/questions/1610030/why-does-flowing-off-the-end-of-a-non-void-function-without-returning-a-value-no

When espanso is built with nixpkgs on aarch64-darwin (which uses clang by default), this results in `espanso daemon` failing and a `Trace/BPT trap: 5` error (please see my issue above).

If one simply adds `return NULL;`, to this specific function (`espanso-detect/src/mac/native.mm:detect_initialize`) espanso builds and runs as expected.

I don't know why cc-rs doesn't seem to show any version of this warning in its build logs: ` warning: non-void function does not return a value [-Wreturn-type]` -- when I make a toy `void *` function and omit a return, I get a warning when compiling with clang and an error / refuses to compile at all in xcode.

I have a one-line PR ready to fix this specific function, but I'd like to expand it to also include some kind of warning or compile error which should help point me to any other instances of this problem in the codebase so I can fix those as well.

Also, I do *not* know why espanso compiled natively with `cargo` works without this error; perhaps it's using a different C standard than nix? Again, I apologize for now knowing much about c-based languages, and I appreciate any input others can offer.

### Steps to reproduce

```console
$ nix run github:nixos/nixpkgs#espanso -- daemon
09:37:07 [daemon(84858)] [INFO] reading configs from: "/Users/n8henrie/.config/espanso"
[...]

Issue #1742, opened 2023-11-06
Want to become an Espanso maintainer? :)


federico-terzi:
Hi everyone!

As many of you have noticed, my contributions to the project have slowed down significantly in the past year, and I'm really sorry about that. The reason is that I've taken on many commitments both in my professional and personal life, and that limited the amount of bandwidth I could dedicate to Espanso.

The good news is that the Espanso community is still growing, and I want to make sure the project has a bright future ahead. For this reason, I'm asking the community: is there someone willing to join the Espanso maintainers team? (which at the moment, it's just me :D ). 

The activity would involve following:
- Helping people on GitHub and Reddit
- triaging issues
- Reviewing PRs and packages on the Hub
- Contributing to the codebase
- Improving the documentation
- Helping the Espanso community foster in general

Note: you could also just help on a specific area :) For example, if you are not comfortable with coding but would like triaging issues, that would already be a great help!

Note #2: I'm not abandoning the project :) I'll still be there if a critical issue comes up, and I don't exclude the possibility of actively contributing again in the future if my life circumstances allow it

If you are interested, please reach out to me :) https://federicoterzi.com/contact-me

Happy expanding!

mistahBen:
I have virtually no experience with rust but Espanso is a utility that I find almost endlessly useful. Would love to try and help in some way(s), in particular documentation/triage.

I've spent most of my time-in regards to Espanso- trying to answer questions over on Reddit. (Username `30ghosts`)

joh6nn:
I would be glad to triage Github issues and help people here when I'm able. I can also help with documentation, testing, and reviewing packages on the hub.

I don't have much rust experience, but I'm interested in learning and would be glad to contribute code and review PRs as I'm able to.

federico-terzi:
Thank you folks! I appreciate the help :) Are you comfortable in coding other languages? I'm asking because one easy place to get started could be the [Espanso Hub](https://github.com/espanso/hub). We have many packages waiting for review, and while there's some basic CI pipeline to validate them, we could do a lot better in terms of automation :) Let me know if you're interested

joh6nn:
Yep, I'm comfortable with several other languages, just not rust :slightly_smiling_face: Reviewing PRs on the hub is something I can do; I've written some private packages for use at work

anshulxyz:
I'd like to get involved in the project. I have a tiny experience in fixing issues in open source Rust project.

[...]

PR #1946, opened 2024-04-13
Add return value to fix runtime crash on nix + aarch64-darwin

n8henrie:
Absence of a return value in a non-void function can be UB and results in a
runtime crash when compiled via nix on aarch64-darwin.

I expect there will be other instances in the codebase; I'm going to
investigate changing the cc-rs incantation to produce a compile-time error to
help identify and correct other instances of this issue.

https://github.com/espanso/espanso/issues/1945
https://github.com/NixOS/nixpkgs/issues/247162

### DIFF
--- a/espanso-detect/src/mac/native.mm
+++ b/espanso-detect/src/mac/native.mm
@@ -137,6 +137,7 @@ void * detect_initialize(EventCallback callback, InitializeOptions options) {
         }
     }];
   });
+  return nullptr;
 }
 
 OSStatus hotkey_event_handler(EventHandlerCallRef _next, EventRef evt, void *userData)


n8henrie:
Actually, it looks like this *does* produce a warning when building in verbose mode, and it looks like this instance in `espanso-detect` may be the only occurrence of this issue in the codebase as of 4c36a0e38f7ba3796b7302fe147c914a98694b66:

```console
$ git checkout 4c36a0e38f7ba3796b7302fe147c914a98694b66
$ cargo build -vv |& grep non-void
[espanso-detect 0.1.0] cargo:warning=src/mac/native.mm:140:1: warning: non-void function does not return a value [-Wreturn-type]
warning: espanso-detect@0.1.0: src/mac/native.mm:140:1: warning: non-void function does not return a value [-Wreturn-type]
$ echo $?
0
```

n8henrie:
Changed from `NULL` to `nullptr` which seems to match the convention in e.g. `espanso-detect/src/x11/native.cpp`

AucaCoyan:
Wow that's a really really good investigation! Thank you for making the PR 😄 
I'll merge after the green CI 👍🏼

AucaCoyan:
All checks passed, squashing!

n8henrie:
Wonderful, thanks! I'll be anxiously awaiting the next (point) release so I can update the package in nixpkgs with this fix -- until then, I have a local overlay that's working. Such a relief to have espanso working again after several months!

Also, if you're at all familiar with nix, I submitted a PR to the popular home-manager project today to enable the espanso module on Darwin.

AucaCoyan:
Thank you! Yes after the call for help in #1742 me and other guys gathered and we are doing some cleanups. Federico is still the admin of the project and I can't do releases, but over time this will convert into a open organization 😄 .
We don't have a date for the next release, but we'll keep you and the community updated!
Feel free to join the discord https://discord.gg/DFcCNDg7bB if you want to chat or hangout

n8henrie:
Sounds good. Yes, I saw the call for help and reached out, but I think several of you had already jumped in to fill the need! Thanks for helping make sure this project stays healthy.

AucaCoyan:
We need a lot more though, if you want, you can contribute in a "isolated" way with PRs, or we can assign you a role (triager, or something). Up to you! The way you feel more comfortable

n8henrie:
> We need a lot more though

That makes sense -- I'd love to be involved in a more official way, my only hesitation is for want of time, being involved in a few other projects already, and with a very busy last half of 2024 for my family (moving). I'm very enthusiastic about rust and would hope to dive in and contribute more to the codebase, though most of my open source contributions have been more on the devops side of things, esp GitHub Actions, bash / python. I'm MacOS / Linux based so can't be much help on Windows, and I have very limited familiarity with C-based languages (which is why this issue took me so long to sort out!).

Issue #1825, opened 2024-01-26
[Meta] Dependencies


AucaCoyan:
Hi! I'll track the dependencies here, as many need an update, we have to:

first, upgrade the security issues informed by [`cargo-audit`](https://lib.rs/crates/cargo-audit).

<details>

```
    Scanning Cargo.lock for vulnerabilities (372 crate dependencies)
Crate:     bzip2
Version:   0.4.3
Title:     bzip2 Denial of Service (DoS)
Date:      2023-01-09
ID:        RUSTSEC-2023-0004
URL:       https://rustsec.org/advisories/RUSTSEC-2023-0004
Solution:  Upgrade to >=0.4.4
Dependency tree:
bzip2 0.4.3
└── zip 0.5.13
    ├── espanso-package 0.1.0
    │   └── espanso 2.2.1
    └── espanso-modulo 0.1.0
        └── espanso 2.2.1

Crate:     chrono
Version:   0.4.19
Title:     Potential segfault in `localtime_r` invocations
Date:      2020-11-10
ID:        RUSTSEC-2020-0159
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0159
Solution:  Upgrade to >=0.4.20
Dependency tree:
chrono 0.4.19
├── simplelog 0.9.0
│   └── espanso 2.2.1
├── mac-notification-sys 0.3.0
│   └── notify-rust 4.2.2
│       └── espanso-ui 0.1.0
│           └── espanso 2.2.1
└── espanso-render 0.1.0
    └── espanso 2.2.1

Crate:     h2
Version:   0.3.18
Title:     Resource exhaustion vulnerability in h2 may lead to Denial of Service (DoS)
Date:      2024-01-17
ID:        RUSTSEC-2024-0003
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0003
Solution:  Upgrade to ^0.3.24 OR >=0.4.2
Dependency tree:
h2 0.3.18
├── reqwest 0.11.17
│   └── espanso-package 0.1.0
│       └── espanso 2.2.1
└── hyper 0.14.26
    ├── reqwest 0.11.17
    ├── hyper-tls 0.5.0
    │   └── reqwest 0.11.17
    └── hyper-rustls 0.23.2
        └── reqwest 0.11.17

Crate:     remove_dir_all
Version:   0.5.3
Title:     Race Condition Enabling Link Following and Time-of-check Time-of-use (TOCTOU)
Date:      2023-02-24
[...]

PR #1956, opened 2024-04-23
refactor: Use workspace dependencies where practical

n8henrie:
When using a workspace, one can ensure a single version of a dependency is used across
the entire workspace by configuring the version in `[workspace.dependencies]` and then
referring to it by `dependencyname.workspace = true` in the individual `Cargo.toml`s.

See also:
- https://github.com/espanso/espanso/issues/1825#issuecomment-2053690690
- https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace


n8henrie:
NB: There are other dependencies for which this might be a good idea, but which weren't using the exact same version, so I haven't incorporated them yet, but I would like that to be part of this PR where practical.

The following python script highlights a few opportunities:

```python
# parse_deps.py
"""Quick script to find and parse all dependency versions in the workspace."""

import tomllib
from pathlib import Path
from collections import Counter
from pprint import pprint

cargo_tomls = list(p for p in Path(".").glob("**/Cargo.toml") if len(p.parents) > 1)
counter = Counter(
    (name, str(version)) for  p in cargo_tomls for name, version in tomllib.loads(p.read_text())["dependencies"].items() if not ("path" in version or "workspace" in version))

if __name__ == "__main__":
    pprint(counter)
```

```console
$ python parse_deps.py
Counter({('serde_json', '1.0.62'): 4,
         ('fs_extra', '1.3.0'): 3,
         ('crossbeam', '0.8.0'): 2,
         ('walkdir', '2.3.1'): 2,
         ('enum-as-inner', '0.6.0'): 2,
         ('dirs', '3.0.1'): 2,
         ('unicase', '2.6.0'): 1,
         ('serde_json', '1.0.61'): 1,
         ('markdown', '0.3.0'): 1,
         ('html2text', '0.12.0'): 1,
         ('natord', '1.0.9'): 1,
         ('zip', '0.5.13'): 1,
         ('scopeguard', '1.1.0'): 1,
         ('sha2', '0.9.6'): 1,
         ('hex', '0.4.3'): 1,
         ('reqwest', "{'version': '0.11.16', 'features': ['blocking'], 'default-features': False}"): 1,
         ('dunce', '1.0.1'): 1,
         ('ordered-float', '2.0'): 1,
         ('indoc', '1.0.3'): 1,
         ('yaml-rust', "{'version': '0.4.6', 'git': 'https://github.com/federico-terzi/yaml-rust'}"): 1,
         ('path-slash', '0.1.4'): 1,
         ('simplelog', '0.9.0'): 1,
         ('clap', '2.33.3'): 1,
         ('enum-as-inner', '0.3.3'): 1,
         ('log-panics', '2.0.0'): 1,
         ('fs2', '0.4.3'): 1,
         ('dialoguer', '0.8.0'): 1,
         ('colored', '2.0.0'): 1,
         ('notify', '4.0.17'): 1,
         ('opener', '0.5.0'): 1,
         ('sysinfo', '0.28.4'): 1,
         ('chrono', "{'version': '0.4.19', 'features': ['unstable-locales']}"): 1,
         ('rand', '0.8.3'): 1,
         ('sys-locale', '0.1.0'): 1})
```

For example, you can see that we're using `serde_json 1.0.62` in 4 places but `1.0.61` in one place. I'm assuming that this was *unintentional* (and if so is the exact problem that this pattern should prevent going forward).

AucaCoyan:
Excellent! Thank you for doing this, I had it planned and you were faster 💯 
> There are other dependencies for which this might be a good idea, but which weren't using the exact same version

Correct, I knew about that when I updated many dependencies a few months ago. We couldn't make it build for the latest version of those dependencies, that's why we left it as is. 
I think those cases are related to some _other_ old dependency or some old API that was broken over time. This project had a hiatus of at least a year in the best case, but several years in some specific dependencies, that's why everything looks dusty.
Thanks again!

PR #1962, opened 2024-05-04
Update to latest clippy

AucaCoyan:
Rust had a recent release, so I asked clippy if they were new lints on the codebase.
There were some!


AucaCoyan:
Good! All green

PR #1967, opened 2024-05-13
Support non-ASCII characters in triggers on macOS

mthies-unibi:
Reports the number of bytes in the UTF-8 encoded character input to the Rust code, as expected there. Formerly the number of Unicode characters/codepoints was reported instead, which did only work for 7-bit ASCII characters.


AucaCoyan:
Thank you! This is a really really prioritized bug in macOS. You've done that much for espanso!

PR #1989, opened 2024-06-13
Word separators

smeech:
I realised we omitted some word_separators that may occur BEFORE words, so have added "( ", "{",  "[", and "<", as well as the additional punctuation marks ":", and ";". The documentation will have to follow suit.

There is a rationale to adding "-" and "_", but I think these can be left to users who need them. 

Minor tidying of security.md.

I think this is straightforward enough to proceed to merge but prefer someone else cast an eye over it first.


AucaCoyan (approved):

AucaCoyan:
All good!

smeech:
Thanks!

PR #1795, opened 2023-12-19
:arrow_up: update `clap`

AucaCoyan:
WIP

this fixes #1792

PR #2009, opened 2024-07-10
:sparkles: try again with clap 3.0

AucaCoyan:
there was an attempt in #1795


AucaCoyan:
Hi @n8henrie Nathan, sorry to ping you, would you mind reviewing this PR?
CI marks all green, but I wanted to consult with another another big brain 😄

espanso/Cargo.toml
```diff
@@ -55,7 +55,7 @@ simplelog = "0.9.0"
 log.workspace = true
 anyhow.workspace = true
 thiserror.workspace = true
-clap = "2.33.3"
+clap = "3.0.0"
```
n8henrie: I would recommend loosening the version restriction to "3" -- no good reason to tie to a minor or bugfix release IMO.

n8henrie: Also, did you look at upgrading directly to clap 4? It looks like v4 was released way back in 2022...

n8henrie (changes requested):

n8henrie:
> sorry to ping you

Happy to chime in :)

espanso/Cargo.toml
AucaCoyan: Yes! I'm on my way. The upgrade guide says it first needs to be updated on 3.0, and then the latest 3.*.
Apparently there are a couple of broken apis since version 3.0.0.
I would also like to use the derive API, is _way_ less verbose and understandable that the current build API

AucaCoyan:
Thanks for the review ❤️

espanso/Cargo.toml
AucaCoyan: The guides [from v2 to v3](https://github.com/clap-rs/clap/blob/v4.4.11/CHANGELOG.md#migrating-1) and from [v3 to v4](https://github.com/clap-rs/clap/blob/v4.4.11/CHANGELOG.md#migrating)

n8henrie: Interesting.

I've used the derive api in a few small personal projects -- pretty nice to work with for the 90% use case!

AucaCoyan: I did the `"3"` change and still compiled, so I'm merging this and starting over for clap v4. Thanks again for the time!

PR #2063, opened 2024-09-17
Resolve issue with cursor placement when using multi-byte characters

### LINKED ISSUE #2062: issue with cursor placement when using multi-byte characters
### Description

When usingthe cursor hints feature, the cursor placement is incorrect if the replace string contains non-ASCII characters (e.g., some emojis). The issue arises because the number of backward steps for cursor positioning is calculated based on the number of chars instead of grapheme clusters. This leads to misalignment between the visual cursor position and the intended location, especially when the replace string includes characters that consist of multiple code units.

_I've already fixed this bug and will create a pull request referencing this bug report_

### Steps to reproduce

Create a trigger with cursor hints that contains non-ASCII characters made up from more than one char, for example ⏱️ or 🕰
```
  - trigger: "test1"
    replace: "[🏢::$|$] [⏱️"
```
This will result in the following `[🏢:|:] [⏱️::]` (where the pipe represents the final cursor position)
This is because ⏱️ counts as two chars, and there fore the number of left steps needed is miscalculated.

### Expected behavior

Expected final cursor position should be `[🏢::|] [⏱️::]`

### Screenshots

_No response_

### Logs

_No response_

### Your environment

- Windows 11
- Espanso version: 2.2.1

### Self-service

- [X] I'd be willing to fix this bug myself.

ndit-dev:
fixes #2062
When using non-ASCII characters in the replace string, the cursor placement is incorrect because the number of backward steps is calculated based on the number of chars instead of graphemes. Some emojis consist of more than one char, for example "⏱️".

**Solution**
Updated the code to count the number of graphemes instead of chars, ensuring correct cursor placement with multi-byte characters.

**Testing**
Ran all existing tests and also compiled it for windows and verified that it now calculates the number of steps correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -786,6 +786,7 @@ dependencies = [
  "log",
  "markdown",
  "thiserror",
+ "unicode-segmentation",
 ]
 
 [[package]]
@@ -3168,9 +3169,9 @@ dependencies = [
 
 [[package]]
 name = "unicode-segmentation"
-version = "1.7.1"
+version = "1.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb0d2e7be6ae3a5fa87eed5fb451aff96f2573d2694942e40543ae0bbe19c796"
+checksum = "f6ccf251212114b54433ec949fd6a7841275f9ada20dddd2f29e9ceea4501493"
 
 [[package]]
 name = "unicode-width"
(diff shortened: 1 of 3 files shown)


AucaCoyan:
Thank you! I already see the benefit without running the tests.
Let them run for a bit to see if we are all okay 😊

AucaCoyan:
I fixed the clippy issue, would you please pull from `dev` again? 🙏🏼

ndit-dev:
Thank you, AucaCoyan! I've pulled in the Clippy fixes and synced my branch. Let's see if the tests pass now! 🔥

AucaCoyan:
Thank you!

Issue #2038, opened 2024-08-15
Delay when sending key strokes


stebberg:
### Description

If I have a windows form with 3 textfields, you can usually fill it out by pressing TAB between the values.
But if I try this in Espanso eg `A\tB\tC` it will all end up in the first (focused) textbox.
In the text expander tool I've been using so far I can insert delays in the snippet, like `A\t(delay 50 ms)B\t(delay 50 ms)C`. Not sure what is going on here but that seems to make Windows to have the time and be able to handle it and actually move focus to the next textbox.

### Motivation

If this wont work you need to have 3 snippets for 1 form.

### Have you tried building it? Or would you like it?

- [ ] I'd be willing to contribute this feature to Espanso myself.

smeech:
Some forms will process the data fast enough for `\t` to move between fields, but others won't, which means you need a separate trigger for each field.

Espanso has no facility to introduce individual delays in the way you have suggested as `replace:` takes place as a single event (believe me, I've tried all sorts of things!), but if the form will work using the `inject` backend, you could try increasing [`inject_delay`](https://espanso.org/docs/configuration/options/) a little.

stewcam:
I have tested the following solution in Windows 11 on MS Forms and it works (but does require PowerShell). I had to add a terminal tab for the C to register in the third edit field.
```
  - trigger: ':fm'
    replace: '{{output}}'
    vars:
      - name: output
        type: script
        params:
          args:
            - C:\Program Files\PowerShell\7\pwsh.exe
            - -Command
            - |
              Add-Type -AssemblyName System.Windows.Forms
              [System.Windows.Forms.SendKeys]::SendWait("{BACKSPACE}{BACKSPACE}{BACKSPACE}A")
              Start-Sleep -Milliseconds 50     
              [System.Windows.Forms.SendKeys]::SendWait("{TAB}")
              Start-Sleep -Milliseconds 50    
              [System.Windows.Forms.SendKeys]::SendWait("B")
              Start-Sleep -Milliseconds 50     
              [System.Windows.Forms.SendKeys]::SendWait("{TAB}")
              Start-Sleep -Milliseconds 50 
              [System.Windows.Forms.SendKeys]::SendWait("C")
              Start-Sleep -Milliseconds 50     
              [System.Windows.Forms.SendKeys]::SendWait("{TAB}")
```

smeech:
I've done similar things with `xdotool` (Linux).
[...]

PR #2067, opened 2024-09-19
Fix delay between keystrokes on Mac.

0x3333:
In the mac implementation, the keystrokes doesn't have the same delay as in other platforms.

The injection is made in chunks of 20 chars. This fixes that, respecting the `inject_delay` config.

Fix the Mac issue in #2038


0x3333:
I'm thinking in changing the behavior. What if the delay is 0, we keep the current code(Which injects ultrafast), if there is a delay, then delay between keystrokes. What do you guys think?

0x3333:
The current implementation will keep the default behavior in case the delay is the same as the default one. If a different than the default delay is provided, it will be issued between keystrokes. I believe this way will not disrupt current users.

AucaCoyan:
Thank you!

PR #2085, opened 2024-09-28
remove 2048 byte clipboard limit for windows

kpym:
This resolves #1612 and #2051 on Windows. The previously hardcoded 2048-character clipboard size is now replaced with an adaptive size based on the actual clipboard content length.

Issue #1675, opened 2023-08-11
Memory leak on Mac


eugenesvk:
**Describe the bug**
After some time Espanso shows up as using ~500Mb memory while a newly launched app has 3 instances with ~20M each

**To Reproduce**
1. Launch the app
2. Wait for a few days(?)

**Expected behavior**
The memory to stay more or less consistent regardless of the time that has passed

**Screenshots**
None

**Logs**
Will run next time the app reaches high memory use. Though for a new app instance the log shows nothing re. memory. Is there some other source of info on what app is using memory for?

**Setup information**
- OS: Mac 13.4.1 Ventura
- Version: espanso 2.1.8

nicklutz15:
Yup, came here after noticing the same (600+Mb). All three back down to ~10mb after $ espanso restart. Will also check log info next time I notice. And personally:

Setup information

OS: macos 12.6.6 monterey
Version: espanso 2.1.8

federico-terzi:
Hey folks!

Thanks for reporting, I'm investigating. Something strange is that while the "Memory" tab usage becomes quite high, real memory is quite stable and low:

<img width="663" alt="image" src="https://github.com/espanso/espanso/assets/20475953/7158e86c-d933-4582-bfcd-bd71191dc0f8">

I'll keep you posted when I find something

federico-terzi:
I've tracked it down to this block https://github.com/espanso/espanso/blob/dev/espanso-detect/src/mac/native.mm#L62 , but I couldn't manage to find a reliable solution to the problem (I'm starting to believe it might be a macOS API bug). I've opened a question on SO, hopefully someone with more experience on Cocoa/Objective C experience can help us figure it out: https://stackoverflow.com/questions/76893657/memory-leak-with-cocoas-addglobalmonitorforeventsmatchingmask

eugenesvk:
Nice find! Meanwhile, is it possible to automatically and silently restart espanso once it reaches a certain memory threshold?
Or at least restart periodically?
I've tried to launch via a plist the `/Espanso.app/Contents/MacOS/espanso` binary, but that doesn't work as for some reason it needs to set up some service instead of launching the app on `start` command
Using `/usr/bin/open .../Espanso.app` worked to start Espanso, but didn't work to quit it (I guess it quits `open` instead?)
Or maybe I'm missing something simpler, here is the plist that I've tried
(as a side note - would be nice to disable the notificaion on launch, it's really not necessary)

[...]

PR #2087, opened 2024-09-29
remove 2048 byte clipboard limit for macos

### LINKED ISSUE #1612: MacOS - Clipboard seem to be limited to 2048 characters
Hello

**Describe the bug**
The clipboard seems to be limited to 2048 characters on MacOS Ventura

**To Reproduce**

I use this code for example : 
```
  # Inserts a markdown code block
  - triggers: [":code", "/code"]
    replace: "```\n{{clipboard}}\n```"
    vars:
    - name: "clipboard"
      type: "clipboard"
```

1. Copy long text
2. Paste with espanso replace

**Expected behavior**
Paste all text

**Logs**
```
10:44:56 [daemon(24085)] [INFO] configuration change detected, restarting worker process...
10:44:56 [worker(24332)] [INFO] engine eventloop has terminated, propagating exit event...
10:44:56 [worker(24332)] [INFO] waiting for engine exit mode...
10:44:56 [worker(24332)] [INFO] exiting worker process...
10:44:56 [daemon(24085)] [INFO] spawning the worker process...
10:44:56 [worker(24556)] [INFO] reading configs from: "/Users/papamica/Library/Application Support/espanso"
10:44:56 [worker(24556)] [INFO] reading packages from: "/Users/papamica/Library/Application Support/espanso/match/packages"
10:44:56 [worker(24556)] [INFO] using runtime dir: "/Users/papamica/Library/Caches/espanso"
10:44:56 [worker(24556)] [INFO] system info: Darwin v13.3.1 - kernel: 22.4.0
10:44:56 [worker(24556)] [INFO] using CocoaAppInfoProvider
10:44:56 [worker(24556)] [INFO] using CocoaSource
10:44:56 [worker(24556)] [INFO] using MacInjector
10:44:56 [worker(24556)] [INFO] binded to IPC unix socket: /Users/papamica/Library/Caches/espanso/espansoworkerv2.sock
10:44:56 [worker(24556)] [INFO] using CocoaClipboard
10:44:56 [worker(24556)] [INFO] monitoring the status of the daemon process
10:44:56 [worker(24556)] [INFO] monitoring the status of secure input
```
**Setup information**
- OS: MacOS (Intel) Ventura
- Version: espanso 2.1.8

Have a nice day

kpym:
This PR aims to resolve #1612 and #2051 on macOS by replacing the previously hardcoded 2048-character clipboard size with an adaptive size based on the actual clipboard content length.

**Note: This pull request is untested on macOS.**  
Unfortunately, I do not have access to a macOS environment at the moment, but I believe this solution should address the issue. Testing is necessary before merging.

This change is similar to the fix in #2085, with the key difference being in the native code for [`clipboard_get_length`](https://github.com/kpym/espanso/blob/895444e7131e6c1066cbf662da3011bfb2b00039/espanso-clipboard/src/cocoa/native.mm#L25-L36). The logic mirrors that of `clipboard_get_text`, except it uses `strlen` instead of `strncpy` to determine the required length.

Please ensure that the code is tested on macOS before proceeding with the merge.

### DIFF
--- a/espanso-clipboard/src/cocoa/ffi.rs
+++ b/espanso-clipboard/src/cocoa/ffi.rs
@@ -21,6 +21,7 @@ use std::os::raw::c_char;
 
 #[link(name = "espansoclipboard", kind = "static")]
 extern "C" {
+  pub fn clipboard_get_length() -> i32;
   pub fn clipboard_get_text(buffer: *mut c_char, buffer_size: i32) -> i32;
   pub fn clipboard_set_text(text: *const c_char) -> i32;
   pub fn clipboard_set_image(image_path: *const c_char) -> i32;
--- a/espanso-clipboard/src/cocoa/mod.rs
+++ b/espanso-clipboard/src/cocoa/mod.rs
@@ -45,7 +45,7 @@ impl Clipboard for CocoaClipboard {
       return None;
     }
     // allocate the buffer
-    let mut buffer: Vec<i8> = vec![0; length as usize];
+    let mut buffer: Vec<i8> = vec![0; length as usize + 1];
     let native_result =
       unsafe { ffi::clipboard_get_text(buffer.as_mut_ptr(), (buffer.len() - 1) as i32) };
     if native_result > 0 {
--- a/espanso-clipboard/src/cocoa/mod.rs
+++ b/espanso-clipboard/src/cocoa/mod.rs
@@ -45,7 +45,7 @@ impl Clipboard for CocoaClipboard {
       return None;
     }
     // allocate the buffer
-    let mut buffer: Vec<u16> = vec![0; length as usize];
+    let mut buffer: Vec<i8> = vec![0; length as usize];
     let native_result =
       unsafe { ffi::clipboard_get_text(buffer.as_mut_ptr(), (buffer.len() - 1) as i32) };
     if native_result > 0 {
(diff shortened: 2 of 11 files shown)


AucaCoyan:
Thanks! this is very much needed! 💯

smeech:
I marked this as Draft, pending macOS etc. testing.

Should this branch become `Main` so that testing can take place with a `Dev` branch? The `Main` branch is *very* out of date, and I'm not sure it serves a useful purpose?

woerndl:
I can confirm this fix works on macOS 14.6.1, also tested with larger chunks of text and edge cases [like this](https://github.com/bits/UTF-8-Unicode-Test-Documents/blob/master/UTF-8_sequence_unseparated/utf8_sequence_0-0x10ffff_assigned_including-unprintable-asis_unseparated.txt).

My attempts to use less memory, explicitly use autoreleasepool and to run on the main thread didn't improve stability or memory usage. Virtual memory fills up with this solution, just as it does with mine. Real and private memory are released with this solution, just as with mine.

AucaCoyan:
Good! I see this good in general sense. I have only one question, is this PR opening a bug about memory or the bug already exists? This is what is not clear to me (sorry for being so out of the loop!)

> My attempts to use less memory, explicitly use autoreleasepool and to run on the main thread didn't improve stability or memory usage. Virtual memory fills up with this solution, just as it does with mine. Real and private memory are released with this solution, just as with mine.

woerndl:
The bug, or more likely bugs, already exist: #1675. I’d say it is not solely related to the keystrokes, as mentioned by Federico in the issue. When you copy and paste large clipboards (which you can now do, thanks to this PR), you can see the virtual memory growing with each repetition, and it doesn't get freed correctly.

woerndl:
I can confirm this also works in macOS 15.0.1.

AucaCoyan:
Superseded by #2089 . Just a coincidence that were two very similar approaches 🤞🏼 .
Thank you very much for your effort!

PR #2101, opened 2024-10-18
⚗️ add support to detect wayland compositor in use

morgan-dgk:
The fragmented nature of Wayland means that there is currently no common way to programmatically retrieve app info for the currently focused app (or to get the current keyboard layout). See PR #1062 for previous discussion on this issue.

This PR adds experimental support to detect the Wayland compositor in use. Currently, this information is just logged but it could be used to decide which standalone executable should be called to get keyboard layout or app info from the compositor.  

See [espanso-appinfo-hypr](https://github.com/morgan-dgk/espanso-appinfo-hypr) as a proof-of-concept for such a plugin to retrieve app info under Hyprland.


morgan-dgk:
Clippy complaining about unused code from `wayland.rs`. I am not super familiar with Rust project structure and unsure why since these functions are called in the same module.

AucaCoyan:
It was some shenanigans with os macros and the like, I did a few fixes and apparently is ready!.
Check it out and let me know if you are okay with it 😄

morgan-dgk:
Looks fine to me @AucaCoyan, thanks for your help.

PR #2027, opened 2024-07-28
📝  improve contributing documentation

AucaCoyan:
We have nowadays two documents to guide new and current contributors how to help the project, `Contributing.md` and `Compilation.md`.
This PR has the objective to merge those two in one single file

PR #2104, opened 2024-10-20
 📝 improve contributing documentation 

AucaCoyan:
We have nowadays two documents to guide new and current contributors how to help the project, Contributing.md and Compilation.md.
This PR has the objective to merge those two in one single file

(I mistakenly closed #2027 )


smeech:
Does this affect the compilation instructions in the docs? Should we be maintaining both or does that risk discrepancies arising between them?

AucaCoyan:
My idea is to move all the Compilation.md into Contributing.md, and only have in Compilation a link to the corresponding section of contributing (just if somebody clicks on a old link going to that file).
[like this](https://github.com/AucaCoyan/espanso/blob/upgrade-contributing/Compilation.md).

This way we have all the info for contributing, and compilation under a better context

smeech:
Yes, I think that's reasonable.

Perhaps we also need links from the docs, for those looking for additional information. Currently, the Windows installation doesn't even mention compilation. On the other hand, there's an argument for not overloading the user documentation, and just leave the GitHub pages for enthusiasts to find.

AucaCoyan:
After some thought I think it would be best to open a section in the documentation to developers.

On a first glance, I thought that it would be easier to have not only the compilation instructions but everything a contributor needs all in one markdown file, under the `espanso` repo. When I finished the document I realized it's a pretty long script.

It's an extra effort to have a documentation in a browser, but I think it's nearly impossible to develop espanso without reading the documentation at all

So we can complete the two objectives in one if we add a development section in the docs. What do you think?

smeech:
I think my view above is unchanged. The docs cater to users, the github readme, to developers. They're two different audiences.

AucaCoyan:
I still stand on the documentation page for everyone, both users and developers (under different sections for each type of user, of course).

Nonetheless I understand the difference between a "simple" user documentation and the implementation specifications. In most scenarios this is usually a talk between two developers. I gave it some thought and I think I can be okay with a [mdbook like `cxx` does](https://github.com/dtolnay/cxx/tree/e247ca05841d2669b39e07e084f1e0079b71fa2b/book). It's way easier to fix than re-deploying the espanso.org site every time we expand on internal documentation (with the network roundtrip traffic for everyone who wants to contribute and more).

I have to refactor a bunch of code then 😁

CONTRIBUTING.md
```diff
@@ -24,7 +41,177 @@ This is a list of things we would like to have and mantain across time. Please d
...
+### Linux
+
+Espanso on Linux comes in two different flavors: one for X11 and one for Wayland.
+If you don't know which one to choose, follow [these steps to determine which one you are running](https://unix.stackexchange.com/a/325972).
+
+#### Necessary dependencies
+
+If compiling on a version of Ubuntu X11 before 22.04 (including 22.04):
+
+```bash
```
smeech: Do we need both of these, as the difference between them could be encapsulated by `libwxgtk3.*-dev`?

smeech (commented):
That all makes sense.

I think the Contents section should be _expanded_ at first glance - it's short, and I overlooked it on first read-through, and the PR section doesn't easily follow the Compilation section without it.

Once merged, I think I should link to the document from the docs, although it would require two additions to https://espanso.org/docs/install/linux/, plus new mentions in https://espanso.org/docs/install/mac/ and https://espanso.org/docs/install/win/.

AucaCoyan:
Yup. This new form of documentation triggers a new way of thinking: What do we write for everyone (users of the app) and what is considered "technical" documentation?
I believe how to install the app from source should be on espanso.org, and mentioned in the mdbook with a hyperlink.
Specifications like windows-portable, non-GUI and other variations live in the dev-docs, as the user probably won't need that level of specific instructions. The shorter the better!

AucaCoyan:
It's ready to review!

smeech (commented):
That all makes sense. I haven't attempted to "buiild the book", however.

I have corrected a couple of typos in https://github.com/espanso/espanso/pull/2104/files#diff-5eee5bef90d4b33e6fbfaf7d89e3f9fa91fb3fe909211f2f2043891c23f2b351 and couldn't resist the temptation to capitalise the proper noun "Rust" although I see that puts me out of line with the occurrences in other files.

AucaCoyan:
Haha, don't worry! I checked and there are no missing lowercase `rust` words (excluding the commandline lines)

AucaCoyan:
@smeech Whenever you say, I'll land this PR 🙌🏼

smeech:
> @smeech Whenever you say, I'll land this PR 🙌🏼

Yes, go for it.

smeech:
Where's `./theme`?
And should the two instances of "(more info here)" link to something?

AucaCoyan:
Whoops, those are things I did in the process, and decided to roll back after I saw how little use for us were.

This is like when you print your final project at school and you realize you made a typo in the front page 🤣 . 
I'll fix it quickly

PR #2127, opened 2024-11-06
Add question in template issues

AucaCoyan:
Currently when you create an issue, it shows this dashboard:
image

But people is confused when they only have a question, they probably don't think it's a bug, but they don't know if a thing is possible, or if it is, how to do it.

That's why I edited the line of Discord to redirect questions to Discord. We have the link in a bunch of places, so it would be weird to click the button `New Issue` and read `Join the discord` inside (not an issue type).

If we change the sentence to: we respond any questions in our discord! we solve this problem of users not knowing about the software, not because it's a bug of espanso (even if it could be after a discussion!)

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,11 +1,11 @@
 blank_issues_enabled: false
 contact_links:
+  - name: ❓ Question
+    url: https://discord.gg/DFcCNDg7bB
+    about: Do you have a question instead of an issue? Meet us in the Discord!
   - name: 📖 Espanso Documentation
     url: https://espanso.org/docs/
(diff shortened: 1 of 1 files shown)


smeech:
I agree. The "bug" template is a bit heavy-going for other Issues - for my "Schema" suggestions, I had to go back in and remove the irrelevant sections.

AucaCoyan:
> The "bug" template is a bit heavy-going for other Issues - for my "Schema" suggestions

Not at all! Bug can be any type of misbehaving of the app. Even a simple thing like missing field on a schema. I find classifying most of the *problems of the users* as bugs first results on a safer operation of the project, and after some triaging determine if it's solvable, relevant o even doable.
It's like putting the worse-case-scenario tag first, and then deescalete from that

AucaCoyan:
For the record the result is:
image

PR #2155, opened 2024-11-26
Include a Search Bar shortcut configuration section in the default config file

olitreadwell:
Include a Search Bar shortcut configuration section in the default config file


smeech:
I think that's useful, especially as most Linux users will need to reconfigure the default _away_ from ALT+SPACE.

Is this an opportunity to add a [`# search_trigger:`](https://espanso.org/docs/configuration/options/#customizing-the-search-trigger) too? I would favour `#search#` rather than that suggested in the docs, and would be happy to change them accordingly.

AucaCoyan:
I'm fine with adding this comments in the `default.yml`
What do you think @smeech ?

AucaCoyan (approved):

smeech:
> I'm fine with adding this comments in the `default.yml` What do you think @smeech ?

Yes, good idea. See my comments above.

AucaCoyan:
Great! @olitreadwell can you do us a favour and add the requested changes? 🙏🏼 
Thank you!

smeech (commented):
Reviewed.

olitreadwell:
Thanks!

I will get back to this later this week and update the pull request with the requested changes.

PR #2171, opened 2024-12-07
fix `clippy` lints

AucaCoyan:


AucaCoyan:
I was about to merge `$auto` when I decided to do first a PR for the clippy lints and then another PR with the `$auto`, so it's easier to read in future tracebacks

PR #2196, opened 2025-01-14
Removed CI workaround and fixed compilation on macOS ARM

jpalharini:
Apparently whatever required multiple configuration changes to the compilation of wxWidgets is no longer necessary ([previously referred issue](https://github.com/actions/virtual-environments/issues/3288#issuecomment-830207746) is closed).

This has the side-effect of using the system's libpng (rather than wxWidgets' ancient fork) on compilation. That change fixes a compilation issue on ARM MacOS, caused by requiring a header file that is no longer present (fp.h).

### DIFF
--- a/espanso-modulo/build.rs
+++ b/espanso-modulo/build.rs
@@ -180,8 +180,7 @@ fn build_native() {
     arch => panic!("unsupported arch {arch}"),
   };
 
-  let is_arm64_ci = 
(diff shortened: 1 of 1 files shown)


AucaCoyan:
Hi, thanks for the request! I see you closed the PR, did you forgot something? 😮

AucaCoyan:
Thank you! Again, I couldn't done this myself as I'm far from knowledgeable in this topic ❤️

PR #2200, opened 2025-01-20
:adhesive_bandage: fix route of the Compilation doc

AucaCoyan:
Forgot to change the path before merging


MRDGH2821:
I think this needs fixing again

AucaCoyan:
Mmm, works as I intended. It guides to how to build the developer docs and check the contents

What do you thing should be going to? the compiling instructions?

Issue #2199, opened 2025-01-19
cargo workspaces aren't being useful for our project


AucaCoyan:
### Description

After some thought and googling, I realized we are not using [cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) for any particular reason.

### Motivation

- Espanso as a binary is the goal of this project
- we do not publish any sub crate and we don't plan to. 
- [~~workspaces don't accelerate build times~~](https://users.rust-lang.org/t/large-private-single-project-workspace-or-not/67185). ~~You can't really try anything in espanso without building it entirely (because we don't really focus on libraries, as I stated previously)~~ This is false-ish. If you split the crates in very (very) tiny pieces, you can have a performance improvement if you have one of those high-cores CPU. [source](https://www.feldera.com/blog/cutting-down-rust-compile-times-from-30-to-2-minutes-with-one-thousand-crates). It's a pain to work though!
- In a monolithic crate, you can test any submodule just like you can test a single crate in a workspace.
- We also don't have different dependencies for different crates, or different flags for the same crate, for example:

```toml
serde = { version = "1", default-features = false }
# and, in another crate
serde = { version = "1", features = ["std"] }
```

- we already did the job to compile all the deps into `[workspace.dependencies]`.

So, what's the reason we are doing a workspace? just because they are cool?

### Alternatives

Just move everything under a single crate.

### work done
- [x] Also, it's a good moment to remove the `rustfmt.toml` file and restore a 4 spaced indents like the standard rust guide.
- [x] `espanso-path` was merged into the `espanso` crate

### Have you tried building it? Or would you like it?

- [x] I'd be willing to contribute this feature to Espanso myself.

n8henrie:
> [workspaces don't accelerate build times](https://users.rust-lang.org/t/large-private-single-project-workspace-or-not/67185).

Are we sure? Seems like it may help in some cases: <https://www.feldera.com/blog/cutting-down-rust-compile-times-from-30-to-2-minutes-with-one-thousand-crates>

If memory serves there are at least a few (lengthy, well written, and interesting!) articles on the topic at <https://fasterthanli.me/>

AucaCoyan:
[...]

PR #2286, opened 2025-04-07
:fire: join `espanso-path` to the main binary crate

AucaCoyan:
helps #2199


schemas/matches.json
AucaCoyan, line 1: this should not be commited! it was a mistake

PR #2297, opened 2025-04-12
:memo: Update supported platforms and add the triage methodology

AucaCoyan:
It's time we do a guide on our supported platforms, and the meaning of high / medium / low priorities

- [x] add links to the epics per platform
- [x] add description to triaging issues

Useful sources I read:
[zig platform support](https://ziglang.org/download/0.8.1/release-notes.html#toc-Support-Table)
[rust platform support](https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support.md)


docs/src/ch02-01-platform-tiers.md
```diff
@@ -0,0 +1,39 @@
+# Platform Tiers
+
+We have different tiers to indicate different levels of commitment for each
+architecture + Operative System.
```
smeech: Is it worth mentioning the different levels of implementation between X11 and Wayland desktop environments?

AucaCoyan: Yes! I missed that

AucaCoyan:
Hey! @smeech @joh6nn @Archigos
I finished the first draft. Can you do a review? 😊 Thanks!

joh6nn:
In general this looks good to me, but I'm not sure about completely replacing Reddit with Discord. My preference would be to have both. Reddit has the advantage that questions+answers posted there can be googled; questions we Nsser in Discord are doomed to be asked over and over again because the answers aren't googlable

AucaCoyan:
Right! I added the reddit again 😊.

Are all on board? any further comments?

PR #2325, opened 2025-04-28
Major CI and Build Process Revamp

### LINKED ISSUE #2301: Fix and enhance CI to our current needs
### Description

In #1928 it was pointed out a way to do the signature in the CI.
Now that I have new credentials, we need to change the secrets and improve some parts of the release CI.

For example, we are not signing windows releases anymore, so the lines related to Sectigo signatures should be removed.

To Do:
- [ ] make a list of the needs
- [ ] create a PR and fix them

### Motivation

So I don't have to manually sign in my mac to do a general release

### Alternatives

_No response_

- [x] I'd be willing to contribute this feature to Espanso myself.

n8henrie:
Revamp CI, remove cargo-make and rust-script

Fixes https://github.com/espanso/espanso/issues/2301

- remove dependency on cargo-make and rust-script
- rewrite the respective cargo-make rust-script scripts in bash
  - we already depend on bash for much of the build process
  - these scripts were largely translations of shell into rust to begin
    with
- update the GHA dependencies (e.g. `actions/checkout`) to their current
  versions
- make the macos executable a universal script, obviating the need for
  separate macos-intel and macos-m1 CI jobs
- remove references to macos-m1 (as macos-arm would have been more
  appropriate but is no longer required)
- remove extranous CI build steps; I will re-add if build times explode,
  but worth testing to see if they are really needed


n8henrie:
Looks like I missed a few PRs since my last pull.

n8henrie:
## TODO

- [ ] re-squash after converting the clang format script
- [x] fix the changes to compilation.md that are now in an mdbook
- [ ] test all the new CI artifacts on the appropriate systems (done for macos, help needed esp for windows!)
    - [x] linux wayland x86 -- no relevant build artifacts (#TODO: why not? maybe consider at some point)
    - [x] linux X11 x86
    - [x] do we build for linux aarch64? -- looks like not. #TODO: maybe consider at some point.
    - [x] macos m1
    - [x] macos intel (at least works via rosetta / `arch -x86_64` 
    - [ ] windows portable
    - [ ] windows installer
    - [x] appimage
- [x] fix docker build
- [x] a few things were failing due to clippy lints -- if those are still failing (after the rebase) should they be fixed here or in a separate PR? -- no longer failing
- [ ] add error message if default `espanso_bin` isn't found
- [x] add nix to CI: https://github.com/espanso/espanso/pull/2328
- [ ] test run of prod signing for macos
- [x] remove this branch from the list of branches for which CI always runs
- [ ] remove debugging  / clean `#TODO`s

n8henrie:
One can see the build artifacts from my fork: https://github.com/n8henrie/espanso/actions

Looks like I need to fix the docker build.

.github/workflows/ci.yml
```diff
@@ -7,10 +7,16 @@ on:
     branches:
       - master
       - dev
+      # TODO: remove this
+      - ci-revamp
```
AucaCoyan: Just a reminder to remove this before merging 👌

scripts/build_windows_installer.sh
AucaCoyan, line 1: I would do this entire file on PowerShell 7.
I can do it!

scripts/build_windows_portable.sh
AucaCoyan, line 1: Same, ps1 is the way to go

scripts/build_windows_resources.sh
AucaCoyan, line 1: And this one too!

scripts/sign_windows_exe.sh
AucaCoyan, line 1: This is a file that we still don't know what to do. We will not going to sign with Sectigo, and Signpath is still on diapers, so if we solve this later, it's fine for me.
It would be better to close the scope a little and try to solve windows certificates in another PR 👍

scripts/resources/macos/formula_template.rb
AucaCoyan, line 1: Whoa! love this!

scripts/publish_homebrew_version.sh
```diff
@@ -1,43 +1,48 @@
...
+  echo "Reading artifacts hashes"
+  SHA256=$(awk -F ' ' '{print $1}' target/homebrew/artifacts/Espanso-Mac-Universal.zip.sha256.txt)
 
-echo "Cloning tap repository"
+  echo "Cloning tap repository"
 
-pushd target/homebrew
-git clone git@github.com:espanso/homebrew-espanso.git
+  pushd target/homebrew
+  git clone git@github.com:espanso/homebrew-espanso.git
```
AucaCoyan, line 28: This is another thing, I believe we are not using the tap anymore.
I have very little knowledge of how to upload a package to homebrew. If my memory serves, we are submitting packages directly to the main hombrew repo

Compilation.md
AucaCoyan, line 1: Of course, as you said before, we need to move this into the `docs/` folder

AucaCoyan (commented):
Some comments to initiate a discussion 🤞.
Thank you so much for you effort! This is a thing that I wanted to do for a long time

.github/workflows/ci.yml
n8henrie: Yes, as a habit I tag such things with TODO (so I can easily grep the repo!)

scripts/build_windows_installer.sh
n8henrie: I would love that. PowerShell seems pretty powerful (very nushell-like), but I am deeply entrenched in bash! As powershell seems like the "always-there built-in" bash equivalent for Windows, I think that fully fits with the spirit of this PR!

scripts/build_windows_portable.sh
n8henrie: No worries. These were a pain to get working in bash because I'm doing it all through GHA (don't have a windows machine)! But I wanted to make sure that everything would work okay without cargo-make and rust-script before I made the PR. With that proven, changing to powershell sounds great!

scripts/publish_homebrew_version.sh
n8henrie: I'm no longer using homebrew at all (nix + nix-darwin convert), but I'm happy to get this working as I know this will be a major path for installation on macos!

Compilation.md
n8henrie: Done!

AucaCoyan:
BTW, I want to deprecate (and remove) the prod-release.yml. I want to use create-release-draft.yml, ci.yml and dev-release.yml only.
So it's okay if you don't update it! Let's put the effort in create-release-draft instead

n8henrie:
@AucaCoyan one other point I wanted to mention -- I mostly did a relatively script conversion of cargo-make and rust-script calls to bash.

For example, using GHA logs, one can see what `cargo make` and `rust-script` were doing: https://github.com/espanso/espanso/actions/runs/14601474485/job/40960425112

```
Calling with args: ["build", "--manifest-path", "espanso/Cargo.toml", "-p", "espanso", "--no-default-features", "--features", "modulo native-tls"]
```

For the most part, I translated this directly to:

```console
$ cargo build --manifest-path espanso/Cargo.toml -p espanso --no-default-features --features modulo,native-tls
```
 
or in some cases the `EXEC_PATH` envvar (previously provided by `cargo-make` was also required.

I think in most cases this can be cleaned up / simplified to something like:

```console
$ cargo build
```

(since the default features are `modulo,native-tls`).

In other cases, it will need a little more verbosity to specify e.g. `--no-default-features --features wayland,modulo,vendored-tls`

Are you okay with me cleaning these up to make them a little less verbose?

I think I can also chase down the `EXEC_PATH` variables and a few other remnants that no longer make much sense.

n8henrie:
Should be ready to turn this draft into a PR soon. Just making sure everything passes on my fork.

AucaCoyan:
> Are you okay with me cleaning these up to make them a little less verbose?

Yes please! the simpler, the better. After that it's just a Cargo.toml default features, or a specific build.rs for mac, and so on.
I'll start with Powershell scripts

n8henrie:
Looks like we should remove `set-output` while we're at it: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

AucaCoyan:
I was hacking the installer script and realized it's needed to do build the windows resources (I knew it, just didn't remember).
That is when I was blocked for a little by the `EXEC_PATH` variable. It's setup by the Makefile [here](https://github.com/espanso/espanso/blob/bbdc9245919fb42cec14250c47db1fdb1e97f6fe/Makefile.toml#L9)
It defines if the binary is located in `./target/debug/espanso` or `./target/release/espanso`

Because the build_windows_resources.ps1 can't know if the user wants the debug or the release version of the binary (both can be available paths simultaneously) I prefer passing a flat `-Debug` or `-Release` flag to the script and finish my day early. Keeping it simple is always better. You can pass args in any idiomatic-bash you feel like it!

scripts/sign_windows_exe.sh
n8henrie: In that case, would you be okay leaving `sign_windows_exe.sh` as is (or perhaps with a comment at the top that it is untested and currently may not be functional) and removing `sign_windows_exe.rs`?

I'm hoping to remove all references to `cargo-make` and `rust-script` to avoid confusion, would prefer not to leave `sign_windows_exe.rs` hanging around.

Or I could remove `sign_windows_exe.sh` completely, just thought that it might serve as useful reference material for a PS script.

n8henrie:
> BTW, I want to deprecate (and remove) the prod-release.yml. I want to use create-release-draft.yml, ci.yml and dev-release.yml only. So it's okay if you don't update it! Let's put the effort in create-release-draft instead

Would it be okay for me to remove `prod-release.yml` completely then?

> That is when I was blocked for a little by the EXEC_PATH variable. It's setup by the Makefile [here](https://github.com/espanso/espanso/blob/bbdc9245919fb42cec14250c47db1fdb1e97f6fe/Makefile.toml#L9)

Yes, that's what I meant when I said:

> or in some cases the EXEC_PATH envvar (previously provided by cargo-make was also required.

I was thinking about either changing it to an argument or a more obviously-named env var. It looks like you prefer argument:

> I prefer passing a flat -Debug or -Release flag to the script and finish my day early. 

Sounds good.

Also -- rethinking a bit, it might be wise to move the required `cargo test && cargo build` logic *into* the scripts themselves, particularly as `build.rs` still doesn't support any post-processing (https://github.com/rust-lang/cargo/issues/545). This means that a simple `cargo build` can get us the binary but not do the rest; if we move the `cargo build` parts into the scripts themselves, we should be able to simplify the CI stuff and build instructions significantly. Does that make sense?

n8henrie:
Another reason is that the "fail fast" (`set -Eeuf -o pipefail`) type stuff is *not* set in the default GHA environment, and part of the pain of debugging this is that failing steps that are written directly in the GHA yaml are just ignored unless we specify this everywhere. Much easier (for me) to ensure they are set in the standalone scripts.

n8henrie:
> That is when I was blocked for a little by the EXEC_PATH variable.

Also -- this is set in the PR here: https://github.com/espanso/espanso/pull/2325/files#diff-4aff73f8108c02d61f96b67590c8589ba099c2c540f620cf9742a536e48ea233R95

Does that also work for powershell?

n8henrie:
I think all of the `cargo make` provided environment variables are now gone.

A few scripts that depended on these for e.g. `EXEC_PATH` now accept a positional argument -- EDIT: which is expected to be the path to the espanso binary to be used for the purposes of that script (default: `target/release/espanso`). This provides an easy way to override the binary for `nix` or to use .e.g `target/debug/espanso` or even use a different architecture if desired).

n8henrie:
I feel pretty good about this so far: Screenshot 2025-04-29 at 15 52 13

AucaCoyan:
It's looking really good 👌.
I feel strange without the need of cargo make, but I knew it was the right direction the day I got this maintainer job.

I'm still working on the powershell scripts 👍

n8henrie:
Things are looking pretty good IMO.

On to testing, and if things seem to be working as expected I'll remove the `TODO`s and debug stuff and switch this out of draft mode.

scripts/sign_windows_exe.sh
AucaCoyan: Removed!

AucaCoyan:
Well then! That's a significant step finalizing all the jobs!
It's late now here, I'll get some sleep and I'll review everything tomorrow with a fresh brain. Thank you again!

.github/scripts/fedora/build_rpm.sh
AucaCoyan, line 1: I would revert 7eba5a5f32f178056695bf6489831d257f77eae9 , 
In the past some users requested to be buildable in rpm systems, and I tried to add it to the CI, but the fedora runners aren't common, and CI never completes.
Can we keep it?

A simple `cargo build --release` should do the trick for the line 11

.github/scripts/ubuntu/build_deb.sh
```diff
@@ -1,24 +1,34 @@
...
+  cargo deb --package espanso --variant wayland -- --no-default-features --features modulo,vendored-tls,wayland
 
-echo "Copying to mounted volume"
-cp espanso-debian-* /shared
+  popd
+
+  find ./espanso/target/debian -name 'espanso_*.deb' -exec cp {} espanso-debian-x11-amd64.deb \; -quit
+  sha256sum espanso-debian-x11-amd64.deb > espanso-debian-x11-amd64-sha256.txt
+
+  find ./espanso/target/debian -name 'espanso-wayland*.deb' -exec cp {} espanso-debian-wayland-amd64.deb \; -quit
```
AucaCoyan, line 26: Just curious, what benefit does `find` in comparison to a `cp`? I honestly don't know

.github/workflows/dev-release.yml
AucaCoyan, line 1: Just a note: It's allright to delete the fedora commented-out workflow. It wasn't working anyways 🤷🏼‍♂️ .
I do want to keep the script for anyone interested in compiling themselves

docs/src/ch03-02-macos.md
AucaCoyan, line 1: May be a dumb question but: shouldn't we indicate that in order to do the bundle the user should add the rust targets and build for each of those
```
rustup target add aarch64-apple-darwin x86_64-apple-darwin
cargo build \
    -p espanso \
    --target x86_64-apple-darwin \
    --release \
    --no-default-features \
    --features native-tls,modulo
cargo build \
    -p espanso \
    --target aarch64-apple-darwin \
    --release \
    --no-default-features \
    --features native-tls,modulo
bash ./scripts/create_bundle.sh
```

right?

scripts/publish_homebrew_version.sh
AucaCoyan: Mmmm, now that I peek again at it, the prod-ci _also_ pushes to the hombrew cask, and not the main repository.
Let's put this to a side and resolve it in another PR, it's not a problem for today

AucaCoyan (changes requested):
First of all, Thank you with a big T!
It's a really good quality of code, I'm really happy about it 😍.

I left a couple of comments, I hope I'm not too pedantic. Still have to learn how to review PRs in a reasonable manner. Apologies in advance if I cause too much trouble!

I leave you with a question about the code snippets:
Why do you leave a dollar `$` in the code snippets? Like this:
```bash
$ echo 'hello'
```

Those things prevent me from click-coping and pasting 🤔. (It's good in some cases though, especially with sudo stuff).

.github/scripts/fedora/build_rpm.sh
n8henrie: Can do. I saw it was getting removed from CI so removed here as well. I have never used fedora and don't have access to a fedora system, so without CI I don't have a way to test and will just leave it as is.

.github/scripts/ubuntu/build_deb.sh
n8henrie: Great Q! It's actually all about avoiding globs (`*`), whose behavior is a bit unpredictable and can even be dangerous.

I whipped up an example bash script showing how globs can inadvertently pass filenames as flags to your command, potentially turning `rm *` into `rm -rf --no-preserve-root *`: https://github.com/quicksilver/Quicksilver/pull/3051#issuecomment-2839600316

Also the behavior of globs depends on the parent shell settings -- globstar, extglob etc.

That's why it's safest to have your bash scripts set `-f`, which disables globs.

Without globs, it's a little more difficult to manipulate unpredictable (or variable) file paths; thankfully `find` is perfectly matched for this need.

Take a look at my example in that comment, let me know if you have questions!

One (trivial) performance win I try to employ is using `-quit` if I'm only expecting one match, using `-mindepth` and `-maxdepth` where appropriate, and using `-exec ... {}... +` if I'm expecting multiple and am offloading to a command that can accept multiple.

docs/src/ch03-02-macos.md
n8henrie: Maybe! Except that I don't think older intel devices actually have the ability to build for ARM, and I didn't want to exclude them. That's why I gave the ability to specify `espanso_bin` instead -- so they can build the bundle for a single architecture if needed or desired. But by default it will except both.

This also provides an ideal workaround for nix.

I'll add an explanatory error message to all the scripts that can accept `espanso_bin` if the default paths aren't found. How does that sound?

AucaCoyan: Yup! that sounds great!

I did some research and it _is_ possible to build arm on a x86 system. [stackoverflow quetion](https://stackoverflow.com/questions/66849112/how-do-i-cross-compile-a-rust-application-from-macos-x86-to-macos-silicon), but I wouldn't take that path because macOS x86 is getting a little bit old. I don't think many developers would use x86 to develop nowadays

n8henrie: I think that depends on running Big Sur or newer though. I'm talking about older intel machines that may still be running 10.x. (Believe it or not, a surprising number of Quicksilver users we support are running pretty old machines that don't support the newer OSes!)

AucaCoyan: I'm sad to hear that 🙁. It's resolved anyways!

n8henrie:
@AucaCoyan another point of discussion -- how would you feel about following `main` for our github actions (e.g. `actions/checkout@main` instead of `actions/checkout@v4`)?

Similar to following `macos-latest` as opposed to pinning to `macos-whatever`, I generally think it's a *good* idea, because when a breaking change is released, we'll know about it sooner rather than later and can pin to the last working version while we figure out a fix. In contrast, if we pin to e.g. `@v4`, the tendency is to *leave* it there until something breaks due to deprecation, at which point we're kind of under the gun (we have nothing we can quickly pin to while we figure out an upgrade path).

Does that make sense? What is your opinion?

n8henrie:
Sorry, I missed this one:

> I leave you with a question about the code snippets: Why do you leave a dollar $ in the code snippets? Like this:

Yes, it's exactly as you said -- my practice is to put *interactive* snippets (that I anticipate a user will type into their terminal emulator) after either `$` or `#` which by convention tells them whether it is run as their user or as sudo. If you specify the syntax highlighting as `console`, *some* syntax highlighters will present it differently than using e.g. `bash`. Additionally, for interactive commands, it helps clearly differentiate input from output:

    ```console
    $ foo() { echo "stuff here"; }
    $ foo
    stuff here
    ```

```console
$ foo() { echo "stuff here"; }
$ foo
stuff here
```

in contrast, I use `bash` and omit the `$` for content that I expect to be written into a script and run from there (and I don't include the output for these):

    ```bash
    foo() {
        echo stuff here
    }
    foo
    ```

```bash
foo() {
    echo stuff here
}
foo
```

The ease of copying and pasting is a drawback! I'm happy to revert these to `bash` and remove the `$` if you'd like, no problem.

AucaCoyan:
> actions/checkout@main branch

Yup! I agree. I've just check the commit history of the repo and all the commits has ✅ and also it doesn't have much activity either, so I think it's safe enough 👍

AucaCoyan:
Yes! I copy a lot from code snippets 😁. Let's meet half-way, keep the `console` snippet and remove the `$` from the lines. Do you agree?

AucaCoyan:
I'll answer an old comment, I read those in a busy day and didn't had time to answer them appropriately:

> might be wise to move the required cargo test && cargo build logic into the scripts themselves
- [this comment](https://github.com/espanso/espanso/pull/2325#issuecomment-2837152406)
 
Mmm, I've advice against that. In the end it's a preference of how big or how long should be a script (or how much responsability a module or piece of code has). 
It's a convenient step if you follow the "normal and common" path (building the application), but sometimes you want to change a step in the script that makes the windows installer, or you want to build the application but not necessarily have the installer. Those two steps being glue-up results in an additional edit to the script a person needs to make in order to _just_ do the step they need to do. It's time-consuming and annoyance of "the computer not doing what you want to do", because the script is called "build installer", not "compile, test and build the installer".
Again, it's a preference, but for me it's one of those places that DRY doesn't pay you enough to take it. What are we saving? just one or two commands? Nah, not worth it 😁

About `cargo doesn't support any post-processing`: I first thought that was a missing feature, but after I did a check with my night pillow, I realized that is not `cargo` is for, that is what task runners are for (any shell you like, `just`, `Make` and tons more). [`uv` has an interesting issue about it](https://github.com/astral-sh/uv/issues/5903#issuecomment-2813946597). I much rather have cargo do a great cargo job, and let me use whatever tool I want to use (and change it for another one like we are doing in this pr ❤️ )

n8henrie:
> Let's meet half-way, keep the console snippet and remove the $ from the lines. 

Nah, `bash` seems fine -- the syntax highlighting seems to be identical on GitHub and we don't have any interactive output. It's no problem!

EDIT: Actually, `console` is already being used elsewhere -- I'll just keep console I guess? Keep it generic (considering zsh and other shells)?

AucaCoyan:
I never understood the difference between the syntax highlighting `console` and `bash`, I use bash wherever I don't have a language, or it's a bash script just to have any colour at all.
Nushell it's not recognized as a syntax, but if you put `bash` some words light up.
With console I rarely see good results, but I'm acting solely on the experience of how many words light up, not by understanding the problem 😆

n8henrie:
> I'll answer an old comment

After some consideration, I agree. This decouples things to a reasonable degree.

AucaCoyan (approved):

AucaCoyan:
I have to fix windows CI, I'll do it later 👍

n8henrie:
My only ubuntu system is too old, so I wasn't able to test the ubuntu / debian artifacts.

Once that is done and the windows parts are ready, I'll remove the debugging code and convert to a full-fledged PR

AucaCoyan:
Hey! @n8henrie 
I took some timeout. What do you think about moving the new nix workflows into a different pr.
This pr has already grown too much, and I would like to merge, fix all the problems, and then make a new release 😊.

n8henrie:
Sure, no problem.

I'm really excited for you to check out the nixosTest I made though -- brings end-to-end integration testing to a whole new level!

n8henrie:
I think this is pretty ready from my perspective, barring only the Windows stuff you were working on @AucaCoyan!

The current MacOS failure is a clippy lint.

AucaCoyan:
> I'm really excited for you to check out the nixosTest I made though -- brings end-to-end integration testing to a whole new level!

Yeeee! I want to do it, but I have to do a release first 🙏.

Well, apparently after a few commits it works here (passes the CI), but I don't see the GitHub Action running in your fork 🤔. Would you please do an empty or silly commit so you trigger the github CI again please?

AucaCoyan:
Sweet! it passes all successfully 🎉  Thank you @n8henrie  for the colossal effort!
I'll remove the TODO lines and merge this 🚀

n8henrie:
I still think we need to test all the artifacts at least superficially -- for example I was unable to test the Debian / Ubuntu artifacts.

AucaCoyan:
Yeah, we should, but I'm excited to land this so we can fix any problem (probably those in creating a new release).
More changes are coming!

smeech:
I've [fixed](https://github.com/espanso/website/pull/105) the Macos M1 link suffix in https://espanso.org/install/ from .zip to .dmg.

Should I remove the "Intel" button from https://github.com/espanso/website/blob/main/src/pages/install.js and the M1 reference? What should I name the macOS button to?

n8henrie:
`M1` should probably be removed / renamed to `ARM64`, since it applies to all M-series (not just the first gen), anywhere that this is relevant.

In the context of the release artifacts, both of these should probably be changed to "Universal" since we are now providing a universal binary.

AucaCoyan:
Hey all!
Yup, nathan is correct on the naming. Luckly is already done!

Yesterday I had to revert some changes in the website, so currently the button on espanso.org downloads 2.2.1 (very old 😢 ).
After doing a lot of work these 24hs, I could accomplish having a release-able binary. Made [this release PR](https://github.com/espanso/espanso/pull/2462) and soon will trigger the workflow to make the github release.

After that, I'll merge [this other pr](https://github.com/espanso/website/pull/101) that solves the botton and the arch issue (just one button for macos).

That should be it!

Of course if it anything happens, we can move forward and continue fixing bugs (or hotfixing it, if there is the need). But honestly no release since 2.2.1 has been so complete I find enough to move the links of the website.

n8henrie:
cc @AucaCoyan 

<https://github.com/espanso/espanso/pull/2325#issuecomment-2835412688>
> BTW, I want to deprecate (and remove) the prod-release.yml. I want to use create-release-draft.yml, ci.yml and dev-release.yml only.

I should re-implement the MacOS codesigning logic that was removed in this PR. How was that done for the recent 2.3 release? Manually?

Do you want that to continue to be manual? I can build it into CI without much trouble.

AucaCoyan:
Hi! Please do! I will have to do it eventually, but for now the cost of dealing with ~1 signature per month is less than investigating how to sign it in CI and probably make horrible mistakes with my credentials in the process.

Having this prod CI ready would mean that we are able to release after **every bugfix** we do on the dev branch 🙏🏼 .
It would be awesome!

n8henrie:
Well thankfully I should be able to mostly follow the groundwork I've laid for [Quicksilver](https://github.com/quicksilver/Quicksilver), which we've been signing and notarizing automatically for a while now. Will have to take a closer look at how to automate uploads to espanso.org, but starting with signed releases seems like a good first step!

It *does* require putting a (base64 encoded) copy of your signing certificate into GitHub Action Secrets. They seem to be very good at hiding the secrets from *accidentally* appearing in any logs, but obviously if a bad actor were able to run arbitrary code in our CI, I'm sure they could find a way to leak things, so we can never say *zero* risk.

PR #2336, opened 2025-05-11
Detect window for wayland with kdotool (KDE only)

rimrimcat:
Uses [kdotool](https://github.com/jinliu/kdotool) to get window title, class, and exec, allowing to use app-specific filters on Wayland

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -774,6 +774,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "cc",
+ "espanso-package",
  "log",
  "widestring",
 ]
--- a/espanso-info/Cargo.toml
+++ b/espanso-info/Cargo.toml
@@ -12,6 +12,7 @@ wayland = []
 [dependencies]
 log.workspace = true
 anyhow.workspace = true
+espanso-package = { path = "../espanso-package" }
 
 [lints]
 workspace = true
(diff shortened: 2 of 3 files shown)


smeech:
This will be big if it facilitates app-specific configurations in Wayland!

rimrimcat:
It requires installing kdotool first, which is only available for **KDE Wayland**. If non-KDE, not sure how to go about it (sorry for giving false hope to non-KDE users).

smeech:
> It requires installing kdotool first, which is only available for **KDE Wayland**. If non-KDE, not sure how to go about it (sorry for giving false hope to non-KDE users).

It's a start, anyway. :smile:

AucaCoyan (commented):
Hey! thank you for submmiting a PR 😊.
I like this so far! My only concern is that other desktops environments will have a wrong exec name `Missing kdotool!` instead of None.
Would be too much to ask to change it to None?
Maybe we can log if kdotool is not found 🤔. What do you think?
Thaank you again, and sorry for the delay!

rimrimcat:
It does sound better to log the message instead of setting it to AppInfo.
No worries at all, it is just a small change.
Any other adjustments that might be added?

AucaCoyan:
Not at all! this looks really good so far 😊

rimrimcat:
Fixed the warnings, turns out I wasn't using clippy for checking, but it should be fine now

AucaCoyan:
Thank you! I'm running it again 😊

AucaCoyan:
A merge was overdue!
Thank you @rimrimcat !

PR #2369, opened 2025-06-27
Update win.rs for optimize env-path register

Zim-Inn:
Fix the extra char ';' append when the old paths str endswith ';'

### DIFF
--- a/espanso/src/path/win.rs
+++ b/espanso/src/path/win.rs
@@ -31,6 +31,11 @@ pub fn is_espanso_in_path() -> bool {
 
 pub fn add_espanso_to_path(_: bool) -> Result<()> {
     let mut paths = read_paths();
+    if let Some(last) = paths.last_mut() {
+        if last.ends_with(';') {
+            last.pop(); // remove the last ';' for pretty join
+        }
+    }
     let exe_path = std::env::current_exe().expect("unable to obtain exec path");
     let parent_path = exe_path.parent().expect("unable to obtain parent path");
 


AucaCoyan:
Hey! thanks for submitting.
I don't understand the bug you are fixing, in the function
```rust
fn read_paths() -> Vec<String> {
    let path_value = read_user_path_value().unwrap_or_default();
    let paths = path_value.split(';');
    paths.map(String::from).collect()
}
```
it already splits by `;`. Why is it needed again? Maybe a screen capture can help me understand 😊

Zim-Inn:
i write a comment in the commit diff page to explain it.

AucaCoyan:
> i write a comment in the commit diff page to explain it.

Mmmm, I still don't get it. Can you give me an example about how your approach is different?
I tried this in a quick script and the `;` doesn't get through the split:

```rust
fn main() {
    let vector = String::from("one;two;tree;;");
    let result: Vec<&str> = vector.split(';').collect();

    println!("result:\n{result:#?}");
}
```

results in:

```console
result:
[
    "one",
    "two",
    "tree",
    "",
    "",
]
```

No `;` are raised anywhere 🤔

Zim-Inn:
> ```
> result:
> [
>     "one",
>     "two",
>     "tree",
>     "",
>     "",
> ]
> ```
Sorry,  I didn't read the impl of `read_path` when I wrote the commit. 
```
fn read_paths() -> Vec<String> {
    let path_value = read_user_path_value().unwrap_or_default();
    let paths = path_value.split(';');  // so that last str can't endswith ';'
    paths.map(String::from).collect()
}
```

But as your example, consider you push a `four` to the reuslt, and then joined the result by `;`, it get `one;two;tree;;;four` -- redundant `;`.
So we should filter out that empty str after split by `;`.
Maybe your change it in func `read_path` will be better?

In a word,  there is a redundant `;` concat in paths, but I did a inoperative commit. 
You can close this PR and optimize the `read_path` to filter out empty str in the vector.

Issue #2366, opened 2025-06-23
"kdotool missing" log spam


sid-6581:
### Description

My system logs are full of this:

![Image](https://github.com/user-attachments/assets/d8e50822-4376-4acd-9cc1-b102953128ea)

### Steps to reproduce

Run espanso and check system logs.

### Expected behavior

Events that happen multiple times a second shouldn't be logged.

### Your environment

Espanso 2.2.3
KDE Plasma 6.4 (Wayland)
Kernel 6.15

### Self-service

- [ ] I'd be willing to fix this bug myself.

AucaCoyan:
Hey Sid! Yes, this was a recent addition in #2336 
Should be fixed! I'll investigate a way around when you aren't in a KDE-based desktop

rimrimcat:
@AucaCoyan I have tried your [branch](https://github.com/AucaCoyan/espanso/tree/move-kdotool-missing-log-at-the-start), and it does fix the annoying logs (I temporarily removed kdotool for this):

```
18:29:51 [daemon(157845)] [INFO] reading configs from: "/home/rimrim/.config/espanso"
18:29:51 [daemon(157845)] [INFO] reading packages from: "/home/rimrim/.config/espanso/match/packages"
18:29:51 [daemon(157845)] [INFO] using runtime dir: "/home/rimrim/.cache/espanso"
18:29:51 [daemon(157845)] [INFO] system info: Arch Linux vrolling - kernel: 6.15.3-arch1-1
---> 18:29:51 [daemon(157845)] [INFO] kdotool missing or not available for the current wayland DE.
18:29:51 [daemon(157845)] [WARN] keyboard layout watcher couldn't determine active layout.
18:29:51 [daemon(157845)] [INFO] watching for changes in path: "/home/rimrim/.config/espanso"
18:29:51 [daemon(157845)] [INFO] espanso version: 2.2.3
18:29:51 [daemon(157845)] [INFO] spawning the worker process...
18:29:51 [daemon(157845)] [INFO] binded to IPC unix socket: /home/rimrim/.cache/espanso/espansodaemonv2.sock
18:29:51 [worker(157852)] [INFO] reading configs from: "/home/rimrim/.config/espanso"
18:29:51 [worker(157852)] [INFO] reading packages from: "/home/rimrim/.config/espanso/match/packages"
18:29:51 [worker(157852)] [INFO] using runtime dir: "/home/rimrim/.cache/espanso"
[...]

PR #2370, opened 2025-06-28
Add support for Niri wayland compositor

rimrimcat:
Checks the XDG_SESSION_DESKTOP to get the appropriate provider between KDE and niri.

### DIFF
--- a/espanso-info/src/lib.rs
+++ b/espanso-info/src/lib.rs
@@ -31,10 +31,6 @@ mod x11;
 #[cfg(feature = "wayland")]
 mod wayland;
 
-#[cfg(target_os = "linux")]
-#[cfg(feature = "wayland")]
-use std::env;
-
 #[cfg(target_os = "macos")]
 mod cocoa;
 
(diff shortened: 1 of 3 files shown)


rimrimcat:
@AucaCoyan 
works on my Niri session:
`21:07:29 [worker(117137)] [INFO] using WaylandNiriAppInfoProvider`
and KDE session:
`21:25:10 [worker(136642)] [INFO] using WaylandKDEAppInfoProvider`

though I get this error on Niri session (which might be unrelated anyway)
`21:07:41 [worker(117137)] [ERROR] error, wl-paste exited with non-zero exit code`

rimrimcat:
Since we are checking for the current DE, it might be good to move the "kdotool missing" log to the `get_provider` function
if I'm correct, the `get_provider` is only run once, compared to the `main` which is run twice (daemon + worker).

AucaCoyan:
Hey! Sorry for the delayed response 😋.
I didn't knew about niri, it's written in rust, so I'll check that out!

I have mixed feelings over moving `kdotool getactivewindow getwindowclassname` from main.rs to the `get_provider` function. I think I would cause a regression of #2366 that was fixed by #2368 (linking the relevant items).
Nevertheless, I would like much more to merge this and deal with a regression than iterate over a PR that is done and I'm delayed reviewing it 😂.

We can fix whatever it causes 🚀 . I'll take the responsability!

Thank you for submitting the PR. it is also very clear and well written

Issue #2387, opened 2025-07-16
Re-add snap support


stew3254:
### Description

Hey all. I love the tool. I use it pretty expensively in my day-to-day, and I know at least 10 others which use the snap variant as well. However, we noticed that it hasn't been updated in a while. I stumbled across https://github.com/espanso/espanso/issues/1848 where snap support has been removed. Could CI be set up to rebuild and publish these automatically every time a new release is made?

### Motivation

I prefer to consume my packages via snap which can be automatically updated.

### Have you tried building it? Or would you like it?

- [ ] I'd be willing to contribute this feature to Espanso myself.

AucaCoyan:
Hey! We are short on human resources to tackle one more platform/package manager. In the past we decided to remove the platforms we can't support, and we have the AppImage as an alternative to snap. Would that work for you? If it's not the case, what is the reason?

We are accepting PRs if you want to submit the workflow to Snap!

Issue #1793, opened 2023-12-18
Install Wayland DEB on Debian 12 error libwx_gtk3u_html-3.0.so.0


lutfiajib:
Just download latest version Espanso Wayland Debian v.2.2.1 from Github, and got error 

1. Download from Githut using espanso-debian-wayland-amd64.deb
2. sudo apt install ./espanso-debian-wayland-amd64.deb
3. call espanso from console
4. Got this error: 

```
espanso: error while loading shared libraries: libwx_gtk3u_html-3.0.so.0: cannot open shared object file: No such file or directory
```

- OS: Debian 12
- Version: v2.2.1 (latest version)

## Momentary fix

As devs continue to work on this, [this comment](https://github.com/espanso/espanso/issues/1793#issuecomment-1919055522) fixes the problem for the time being.

guillaume-martin:
I have the same bug with espanso-debian-x11-amd64.deb v.2.2.1 (OS: Debian 12)

Rody-Oldenhuis:
Same here on Ubuntu 23.10/Mantic.

FluffMe:
Same here on X11 deb.

Used an AppImage version as a workaround for now. Works with no issues.

Rody-Oldenhuis:
[Building from source](https://espanso.org/docs/install/linux/#wayland-compile) also seems to work OK (using `cargo-make` `v0.37.5` / Ubuntu 23.10)

gavinlaking:
Sadly, the first step of building from source is failing for me on Debian 12 Bookworm and Debian testing as `Unable to locate package libwxgtk3.0-gtk3-dev`.

Edit:

I can however install `libwxgtk3.2-gtk3-dev`, so I did!. Unfortunately, despite compiling again from source, it still fails when running `espanso --version`; with the following error:

```
espanso: error while loading shared libraries: libwx_gtk3u_html-3.0.so.0: cannot open shared object file: No such file or directory
```

Edit 2:

The compiled version appears to run fine (at least `./target/release/espanso --version` doesn't crash), until it is moved or copied to `/usr/local/bin/espanso` via `sudo mv target/release/espanso /usr/local/bin/espanso`. I don't know enough about Cargo to deduce what is actually happening.

Edit 3:

For those triaging, I was just having a look through the issues and noticed this one; https://github.com/espanso/espanso/issues/1739 which is outlines similar issues to this, but no solution either.

zchap26:
[...]

PR #2434, opened 2025-08-12
re-add snap support

slapcat:
Related to #2387.

I've updated the old snapcraft recipe to be based on core24 instead of core18.

There is also a workflow to publish the snap to the candidate and/or edge channel. For this to work, you'll need to add the snap store login info as a repository secret called `STORE_LOGIN`. Instructions can be found [here](https://canonical-robotics.readthedocs-hosted.com/en/latest/how-to-guides/packaging/build-and-publish-snap-with-github-actions/#setting-the-secret).

It appears the current owner of the snap in the store is @federico-terzi: https://snapcraft.io/espanso

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -173,3 +173,21 @@ jobs:
           name: MacOS Artifacts
           path: |
             Espanso-Mac-Universal.zip
+
+  snap:
+    runs-on: ubuntu-latest
+    outputs:
+      snap-file: ${{ steps.build-snap.outputs.snap }}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: snapcore/action-build@v1
+      id: build-snap
+    - run: |
+        sudo snap install --classic --dangerous ${{ steps.build-snap.outputs.snap }}
+    - uses: snapcore/action-publish@v1
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
+      with:
+        snap: ${{ steps.build-snap.outputs.snap }}
+        release: ${{ startsWith(github.ref, 'refs/tags/') && 'candidate' || 'edge'}}
+
--- a/snap/hooks/remove
+++ b/snap/hooks/remove
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+echo "Stopping Espanso..."
+killall espanso
+
+# Here I've also tried to unregister the Systemd service, but couldn't manage to. 
+# The remove hook is run as Root, but the systemd service is registered as a user.
+# I couldn't find any (working) way to unregister a user systemd service as root.
+# I've also tried these solutions: https://unix.stackexchange.com/questions/552922/stop-systemd-user-services-as-root-user
+# but none seemed to work in this case.
+# If you manage to find a way to do so, feel free to open an issue so that
+# we can improve the hook! Thanks
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,112 @@
+name: espanso
+version: 2.2.4
+summary: A Cross-platform Text Expander written in Rust
+description: |
+  espanso is a Cross-platform, Text Expander written in Rust.
+
+  ## What is a Text Expander?
+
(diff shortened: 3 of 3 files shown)


AucaCoyan:
Hey! thank you so much for submitting a PR 😄 . To land this I have some questions. We may merge it anyways and resolve this on the future, but at least those are starting points:

- this is based on the `dev` branch, which is a little more unstable than releases (quarterly). If snap users are okay with that, no problem! otherwise if they want something a bit more stable I would recommend installing as an AppImage or a deb package. Another path we could take is adding this build step into `create-release-draft.yml` that makes everything and then I review the results and publish the release.
- espanso has 2 builds, wayland and x11. The default flags are for x11. Both builds are incompatible with each other. I assume snaps are used everywhere (any flavour of ubuntu), so that would mean that either the snap should have "two versions", or to know which one install based on the desktop environment of the user
- about `snap/hooks/remove`. `espanso service unregister` removes the systemd service 🤔 . I wonder if you can run espanso service unregister before uninstalling it. Does that solve the issue?
- lastly a really dumb question: did you tried the app? 😄  do you like it and worked well?

Thanks!

stew3254:
@AucaCoyan it seems to be working for me on X11 (using awesome-wm). @slapcat had provided the snap to me.

I did have an issue with one of the packages not working within the snap, but after his investigation, it looks like it might be related to [this issue](https://github.com/espanso/espanso/issues/1793).

Regarding the issues mentioned if I'm able to give my two cents:
1. It would be nice if we could have both dev and stable. You can create different snap channels for that.
2. Again, channels could be used to differentiate between X11 and Wayland if they're incompatible
3. Probably
4. Love espanso and use it heavily!

slapcat:
Thanks for looking this over @AucaCoyan!

I resolved the package issue that @stew3254 mentioned, so now the snap can be installed without dependencies and just works on X11. Regarding your other points:

- The workflow I made will publish all changes to the "edge" channel and tagged releases to the "candidate" channel. Stable candidate versions can be manually promoted to the "stable" channel after testing. This is a common snap release cycle, but feel free to adjust if it doesn't suit you. (We may also be able to add a promotion step to `create-release-draft.yml` in order to automate that process.)
- Following the previous snap recipe, this one only builds for X11. I can modify it to build both for X11 and Wayland, but there isn't currently a track for Wayland on the snap store, so we can't release two different builds of the snap currently. Adding a track is non-trivial because a [request needs to be raised and voted on](https://forum.snapcraft.io/t/process-for-aliases-auto-connections-and-tracks/455). This should probably be done by the owner of the snap (the creator of espanso).
- The disclaimer in `snap/hooks/remove` is still valid. The same problem occurs if you try to run `espanso service register` on install. The snap hook runs it as root and it fails. I'd be happy to investigate a workaround to this if snap support is restored.
- Yes! @stew3254, myself, and many of our colleagues rely on espanso everyday. I built a snap of the latest version based on the snapcraft.yaml submitted here and it has been working well the past couple days.

slapcat:
@AucaCoyan Please let me know if you need anything else in order to merge this PR!

AucaCoyan:
Hey Slapcat! Sorry for the delay 😓 . I had a busy time 15 days ago, then macOS decided to stop compiling, and now the release is also not signed, so it's un-runnable from the Applications folder.

I promise that as soon as I get a little rest from this I'll come back with this 💯 . Thank you so much for the patience!

AucaCoyan:
Hey! Sorry again for the delay, there was much turmoil with the few last releases (especially in macOS).
I tasked myself to investigate the snap ecosystem, and submmit a different snapcraft.yaml, so we have normal "production" images in snapcraft instead of dev commit, but I coudn't do it.

Instead of stalling even more the PR, I decided to merge it right away and fix the problem in the future, at least you can benefit from having a development snap image, instead of having nothing at all 😢 .

Again, forgive me for the delay!

slapcat:
@AucaCoyan thanks so much!! This is a big help. I'm happy to help with whatever you need to get the production snap updated. Just let me know.

As mentioned above, I can also update the snapcraft to build snaps for Wayland, but the channels will need to be requested by the maintainer. If you can get those setup, I can update build recipe.

PR #2501, opened 2025-10-04
Fix json label replace (#1702)

### LINKED ISSUE #1702: Label Field Incorrectly Interpreted as Replace Field in espanso match list -j Output
**Describe the bug**
When running the `espanso match list -j` command, the `label` field in my Espanso match configuration is incorrectly interpreted as the `replace` field.

**To Reproduce**
1. Define a match in Espanso with both `replace` and `label` fields, as shown in the example:
    ```
    - trigger: "??ai.tests"
      replace: "As a software developer..."
      label: "Write tests"
    ```
2. Run the `espanso match list -j` command.
3. Observe the output, noting that the `label` field takes the place of the `replace` field in the resulting JSON.

**Expected behavior**
The `label` field should be properly represented as `label` in the output JSON, separate from the `replace` field. The output should look like:
```
{
  "triggers": [
    "??ai.tests"
  ],
  "replace": "As a software developer...",
  "label": "Write tests"
}
```

**Setup information**
- OS: macOS 13.5.1 22G90 arm64
- Version: espanso 2.1.7-beta

gcailly:
The `espanso match list --json` command was incorrectly showing the label value in the replace field when a label was present.

  Now the JSON output correctly shows the actual replacement text in the `replace` field and the label separately in the `label` field.

Fixes #1702

### DIFF
--- a/espanso/src/cli/match_cli/list.rs
+++ b/espanso/src/cli/match_cli/list.rs
@@ -105,9 +105,15 @@ pub fn print_matches_as_json(match_list: &[&Match]) -> Result<()> {
             MatchCause::Regex(regex_cause) => vec![regex_cause.regex.clone()],
         };
 
+        let replace = match &m.effect {
+            espanso_config::matches::MatchEffect::Text(text_effect) => text_effect.replace.clone(),
+            espanso_config::matches::MatchEffect::Image(image_effect) => image_effect.path.clone(),
+            espanso_config::matches::MatchEffect::None => String::new(),
+        };
+
         entries.push(JsonMatchEntry {
             triggers,
-            replace: m.description().to_string(),
+            replace,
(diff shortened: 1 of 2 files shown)


AucaCoyan:
perfect!
Just to know what are the default results, those are:
```json
[
  {
    "triggers": [
      ":espanso"
    ],
    "replace": "Hi there!",
    "label": null
  },
  {
    "triggers": [
      ":date"
    ],
    "replace": "{{mydate}}",
    "label": null
  },
  {
    "triggers": [
      ":shell"
    ],
    "replace": "{{output}}",
    "label": null
  }
]
```

- we are doing `triggers` as an array
- and label is always added as the key, and when you don't have any value, the value is `null`

Those sound strange at first, but its the current implementation 👍🏼

Issue #1766, opened 2023-12-06
Configurable character limit on regex triggers


AucaCoyan:
### Discussed in https://github.com/espanso/espanso/discussions/1253

<div type='discussions-op-text'>

<sup>Originally posted by **Staples1010** January 23, 2022</sup>
I would like to use larger character counts in my arguments, I'm a web dev and would like to use it to create anchors. For example:
```<a href="{{arg1}}" target="_blank">{{arg2}}</a>```
links and anchors will pretty much always be over 30 characters :/

Could it be configurable in the default.yml file?</div>

Hi! Just transferring this discussion into an Issue. First contributors more than welcome!

**Comment by joh6nn** [comment]
I doubt there's a way to accomplish this that doesn't impact performance. I think if people need longer arguments the best option is to use the forms extension

**Comment by AucaCoyan** [comment]
Yeah, that's true. At least we could test some increased limits and see how much longer it takes.

**Comment by smeech** [comment]
No harm in having it as an *option* with appropriate warnings regarding performance.

**Comment by gruns** [comment]
just ran into this issue myself: `replace` strings longer than 103 characters in length fail

103 chars succeeds:
```
- triggers: ['@@grep']
  replace: |
    grep -riI --exclude-dir=node_modules --exclude-dir=dist --exclude-dir=.to --exclude='*.min.js*' '$|$' .
      
>>> len("grep -riI --exclude-dir=node_modules --exclude-dir=dist --exclude-dir=.to --exclude='*.min.js*' '$|$' .")
103
```

104 chars fails:
```
- triggers: ['@@grep']
  replace: |
    grep -riI --exclude-dir=node_modules --exclude-dir=dist --exclude-dir=.tox --exclude='*.min.js*' '$|$' .
      
>>> len("grep -riI --exclude-dir=node_modules --exclude-dir=dist --exclude-dir=.tox --exclude='*.min.js*' '$|$' .")
104
```

`use_simulator: true` doesnt work as a workaround, unfortunately

is there a standard workaround here?

AucaCoyan:
Not that I know yet. I think we need to change the code to be able to do longer regexes.
Of course it's necessary that this would be hidden it under a custom option `regex-length = 300` or something like that.

smeech:
> just ran into this issue myself: `replace` strings longer than 103 characters in length fail
> 
> 103 chars succeeds:
> 
> ```
> - triggers: ['@@grep']
>   replace: |
[...]

PR #2504, opened 2025-10-05
feat: Add max_regex_buffer_size config option

aldair-meneses:
This PR introduces a new configuration option, `max_regex_buffer_size`, allowing users to control the maximum number of characters held in memory for regex trigger matching and solve the issue #1766 .

## Context

Previously, the buffer size for regex matching was hardcoded to 30 characters. This posed a limitation for users wanting to create triggers longer than this size.
This change makes the regex buffer size configurable in the configuration, providing a better balance between capability and resource consumption, tailored to each user's needs.

---

## Changes Made

* **New Configuration Option:** Added `max_regex_buffer_size` to the configuration files, with a default value.
* **Config Trait:** Updated the Config trait and its implementations to include the `max_regex_buffer_size()` method.
* **Engine Integration:** The regex matcher now uses the value from the configuration instead of a hardcoded one.
* **Documentation:** Added comments in both the source code and the default configuration file to explain the new option.

---

## How to Test

1. Open your `default.yml` configuration file.
2. Set a small buffer size:

   ```yml
   max_regex_buffer_size: 10
   ```
3. Create a match with a regex trigger longer than 10 characters:

   ```yml
   matches:
     - regex: ":thisislongerthan10"
       replace: "This should not work."
   ```
4. Save the file and let espanso restart.
5. Try to type `:thisislongerthan10`. The replacement should **not** occur.
6. Now, increase the buffer size in `default.yml`:

   ```yml
   max_regex_buffer_size: 20
   ```
7. Save and let espanso restart.
8. Try to type `:thisislongerthan10` again. The replacement should now work correctly.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "[json]": {
+        "editor.defaultFormatter": "biomejs.biome"
+    },
+    "[jsonc]": {
+        "editor.defaultFormatter": "biomejs.biome"
+    }
+}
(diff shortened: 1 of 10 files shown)


AucaCoyan:
I tested it and it works perfectly!
Thank you so much for the effort, I think it's a contribution of really really good quality ❤️.
It's pending to add `max_regex_buffer_size` into `schemas/config.schema.json` only. I'll add it and merge

PR #2508, opened 2025-10-05
Fix empty triggers causing expansion on every word separator (#1616)

### LINKED ISSUE #1616: Empty trigger expands on word_separator
**Describe the bug**
If a trigger is set to "" (empty), it is expanded by every word_separator. If multiple triggers are set to "" it triggers the search on every separator

**To Reproduce**

```yml
- trigger: ""
  replace: text
- trigger: ""
  replace: text2
```

**Expected behavior**
It should not trigger if the trigger is empty or warn that the config is invalid. (Some edge cases may be able to use this 🤔)

It would be nice to allow an empty trigger for disabling the trigger and quickly find disabled triggers later with a search for `trigger: ""`

**Setup information**

- OS: MacOS 13.4
- Version: 2.1.8 (Homebrew)

gcailly:
When a trigger was set to an empty string (""), it would expand on every word separator, causing unexpected text replacements.

Now empty triggers are filtered out during configuration parsing, and a warning is displayed in the logs to inform that empty triggers are not allowed and have been ignored.

Fixes #1616

### DIFF
--- a/espanso-config/src/matches/group/loader/yaml/mod.rs
+++ b/espanso-config/src/matches/group/loader/yaml/mod.rs
@@ -140,6 +140,24 @@ pub fn try_convert_into_match(
         yaml_match.triggers
     };
 
+    // Filter out empty triggers and add a warning
+    let triggers = if let Some(mut triggers_vec) = triggers {
+        let original_len = triggers_vec.len();
+        triggers_vec.retain(|t| !t.trim().is_empty());
+        if triggers_vec.len() < original_len {
+            warnings.push(anyhow!(
+                "empty triggers are not allowed and have been ignored"
+            ));
+        }
+        if triggers_vec.is_empty() {
+            None
+        } else {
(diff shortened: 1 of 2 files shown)


AucaCoyan:
You sir are lit this weekend 🔥 🔥 
This is also great code!

gcailly:
To be honest, I have only some lower-intermediate Python knowledge and I don’t know Rust, so my skills are quite limited.
I mainly use Claude Code, and I try to be careful, checking the documentation and testing things.
I love Espanso, and I’d be very happy if I could help you, for example by fixing some bugs or reducing the number of open issues. I currently have some time to do this. 🙂

AucaCoyan:
Don't worry! I accept contributions of all levels. So far you did great! doesn't matter from what the tooling.

If you are interested in rust, there is no better place than [the rust book](https://doc.rust-lang.org/stable/book/). There aren't even better paid courses!

PR #2511, opened 2025-10-06
Fix Cmd+Enter not submitting forms on macOS (#2457)

### LINKED ISSUE #2457: Ctrl+Enter for submitting a form with a multi-line text field conflicts with shortcut key for right-click menu on Macs
### Description

I am using Espanso on my MacBook Pro M1. Whenever I use an expansion containing a form with a multi-line text field, pressing Ctrl+Enter as stated to submit the form brings up the right-click menu instead. In order to avoid this shortcut key conflict, it needs to be something else on Macs, such as Cmd+Enter.

Ctrl+Enter works on Windows, however.

### Steps to reproduce

Add the following matches YAML file ([multiline.yml](https://github.com/user-attachments/files/22281392/multiline.yml)):

```yaml
matches:
  - trigger: ":testmultiline"
    replace: "{{f.input}}"
    vars:
      - name: "f"
        type: form
        params:
          layout: "Text:\n[[input]]"
          fields:
            input:
              multiline: true
```

Type `:testmultiline`, then enter something into the form, and try to submit by pressing Ctrl+Enter.

- On Mac, it brings up a context menu as shown in the attached screenshot.
- On Windows, it works just fine by printing the form input's contents.

### Expected behavior

- The shortcut key hint should show Cmd+Enter or some other non-conflicting combination on Macs.
- The form popup should submit upon pressing the stated shortcut key combination.

### Screenshots

Image

### Logs

_No response_

### Your environment

- Operating system and version: Mac OS Sequoia 15.6.1
- Espanso version: 2.2.1

### Self-service

gcailly:
Changed form submission shortcut from WXK_RAW_CONTROL to CmdDown() to support native keyboard shortcuts on all platforms :
- macOS: CMD+Enter (instead of CTRL+Enter which opens context menu)
- Windows/Linux: CTRL+Enter (unchanged)

Also updated the help text to show the correct shortcut for each platform.

Fixes #2457


AucaCoyan:
Currently it doesn't work. Apparently the flag `__WXOSX__` doesn't detect it's macos. I compiled the code in M2 and it still shows the label `(or press CTRL+Enter to submit, ESC to cancel)`, so my guess is that the `#ifndef` is not working correctly.

Looking at the [constants page](https://docs.wxwidgets.org/3.1/page_cppconst.html) I tried with `__WXMAC__` (older version) and couldn't make it work because of a misconfiguration of my machine:
```
$ ./target/debug/espanso start
unable to start service: timed out
Hint: sometimes this happens because another Espanso process is left running for some reason.
      Please try running 'espanso restart' or manually killing all Espanso processes, then try again.
```

I need a bit more to test this before merging, so I'm going to tag this to the next release 👍🏼

AucaCoyan (changes requested):
Future @AucaCoyan , please test it again!

AucaCoyan (approved):

reagle:
This would be handy to have, why did it get stuck? At the least, the Windows only message about "ctrl+enter" in the form should be removed/changed on macOS.

PR #2516, opened 2025-10-06
Fix date extension crash on invalid format (#993)

### LINKED ISSUE #993: Date format doesn't exist, don't crash
**Is your feature request related to a problem? Please describe.**
When you use a percent encoding format that doesn't exist within the date command, it causes espanso to crash with an unhelpful error in the log like so:

```18:51:08 [ERROR] thread 'daemon_background' panicked at 'a Display implementation returned an error unexpectedly: Error': /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/alloc/src/string.rs:2202
stack backtrace:
   0: backtrace::backtrace::trace
   1: backtrace::capture::Backtrace::new
   2: log_panics::init::{{closure}}
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic_handler::{{closure}}
   5: std::sys_common::backtrace::__rust_end_short_backtrace
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::option::expect_none_failed
   9: <espanso::extension::date::DateExtension as espanso::extension::Extension>::calculate
  10: <espanso::render::default::DefaultRenderer as espanso::render::Renderer>::render_match
  11: espanso::engine::Engine<S,C,M,U,R>::inject_match
  12: espanso::matcher::<impl espanso::event::KeyEventReceiver for M>::on_key_event
  13: <espanso::event::manager::DefaultEventManager as espanso::event::manager::EventManager>::eventloop
  14: espanso::worker_background
  15: std::sys_common::backtrace::__rust_begin_short_backtrace
  16: core::ops::function::FnOnce::call_once{{vtable.shim}}
  17: std::sys::unix::thread::Thread::new::thread_start
  18: _pthread_start```

**Describe the solution you'd like**
In this case, it would be useful to try and parse whether the format is correct and if not, throw that into the log. Without crashing of course :P 

**Describe alternatives you've considered**
Looking up the formatting options that exist and using them properly.

gcailly:
Invalid date format specifiers (e.g., `%Y-%m-%dT%H:%i:%s%Q`) caused the daemon to panic.

Changes :
- Add `DateExtensionError` enum with `InvalidFormat` variant
- Use `write!()` to catch formatting errors instead of calling `.to_string()` directly
- Return `ExtensionResult::Error` instead of panicking
- Add regression test with the problematic format from the issue

Fixes #993

### DIFF
--- a/.github/workflows/ci-arm.yml
+++ b/.github/workflows/ci-arm.yml
@@ -15,9 +15,6 @@ jobs:
     runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@main
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          components: clippy, rustfmt
       #   - name: Format
       #     run: cargo fmt --all -- --check
       - name: Install Linux X11 dependencies
@@ -57,9 +54,6 @@ jobs:
     runs-on: windows-11-arm
     steps:
       - uses: actions/checkout@main
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          components: clippy, rustfmt
       #   - name: Format
       #     run: cargo fmt --all -- --check
       - name: Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,9 +30,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@main
-      - uses: dtolnay/rust-toolchain@stable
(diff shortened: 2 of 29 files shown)


AucaCoyan:
Just reverted the mardown part that is for a separate PR 👍🏼

AucaCoyan:
Thank you! this fixes perfectly! 🚀

PR #2518, opened 2025-10-06
Fix XML tags not escaped in search bar (#974)

### LINKED ISSUE #974: Quick filter doesn't display correctly
**Describe the bug**
I have a snippet looks like this:
```
  - trigger: "kjk"
    replace: "<kdb>$|$</kbd>"
```
However, when I search it in the quick filter GUI, it looks like this:
image

Seems the snippet isn't correctly escaped in the filter UI.

**Setup information**
- OS: Windows 10
- Version: espanso 2.1.3-alpha

gcailly:
HTML special characters in snippet labels were being interpreted as HTML tags instead of displayed as text in the search bar.

Now they are displayed :

Capture d'écran 2025-10-06 232552

Capture d'écran 2025-10-06 232529

- Added `EscapeHtml()` helper function in `espanso-modulo/src/sys/search/search.cpp`
- Applied HTML entity escaping (`&` → `&amp;`, `<` → `&lt;`, `>` → `&gt;`, `"` → `&quot;`) to both labels and triggers before rendering.

Fixes #974


AucaCoyan:
I merged #2519 which had these changes patched.
Merged from dev again and we can see there are no pending changes. So I'm closing this PR. 

This counts as a done, not cancelled! 
Thank your your contribution ❤️

Issue #385, opened 2020-07-23
Config files don't support `.yaml`?


4lch4:
This is more of a question than an actual issue. I tried creating some new "packages" under the user config folder and they had the `.yaml` extension as opposed to `.yml`. They don't seem to be getting picked up by espanso though.

When I created a new `social.yaml` and gave it a simple match, it wasn't being picked up by espanso, which leads me to believe the `.yaml` extension isn't supported. I'm not too in touch with YAML but after some Google searches, `.yaml` is the "recommended"[\[0\]][0][\[1\]][1] version and `.yml` is the more popular but equally as valid version.

Here's the content of the mentioned `social.yaml` file:

```yaml
name: 'social'
parent: 'default'

matches:
  - trigger: ":tst"
    replace: "Thisis just a test."
```

Is this intentional or am I just missing something?

[0]: https://en.wikipedia.org/wiki/YAML
[1]: https://yaml.org/faq.html

federico-terzi:
Hey @4lch4,

That's intentional, I wanted to keep it short and followed what some of the most popular tools were doing (such as Docker).
That said, it might be a good idea to support the other format as well, so I'll add it to the roadmap 

Thanks for your help :)

smeech:
Is this still planned?

bannert1337:
I just had this exact problem and was frustrated as to why the expansions I had created were not working. At some point I realised that the file extension was `.yaml` instead of `.yml` and therefore it would not load.

From a user experience perspective, and considering that `.yml` and `.yaml` are widely accepted as interchangeable extensions for YAML files in many other tools and ecosystems, it *feels* like unexpected behavior or an unnecessary limitation. A user reasonably expects either extension to work. Ignoring `.yaml` files silently (without an error message saying "Ignoring file with .yaml extension, use .yml") can lead to frustrating debugging, as I have experienced.

AucaCoyan:
Hey! I need to chime in to say _yes, this is important!_. I remember reading some code somewhere that checks for the extension (sorry for the poor accuracy).
Whatever is causing espanso not recognize `.yaml` extension, should be fixed.

PR #2519, opened 2025-10-07
Add support for .yaml extension for config and match files (#2184)

### LINKED ISSUE #2184: Use the officially recommended extension for files containing (a) YAML document(s)
### Description

The FAQ on yaml.org has stated that the officially recommended extension for files containing YAML documents is `.yaml`, at least since September 2006 (two-thousand and six!).  
There is an (older) YML format which is an XML derivative.

### Steps to reproduce

_No response_

### Expected behavior

Espanso should generate and use `default.yaml` and `base.yaml` out of the box, offering to correct files with YAML documents using the long outdated extension, during transition you could make symlinks from the incorrect name to the correct one (`base.yml` linked to `base.yaml`)  If `default.yaml` cannot be found a warning should be issued and `default.yml` tried as fall-back.
Make sure the documentation is updated to reflect the use of the recommended extension.

### Screenshots

_No response_

### Logs

_No response_

### Your environment

- Public source code:
- Operating system and version (e.g., Ubuntu 20.04.2 LTS):
- Espanso version (e.g., 2.1.4-beta):

### Self-service

gcailly:
This PR adds support for the `.yaml` extension for configuration and match files, while maintaining backwards compatibility with existing `.yml` files.

- New config and match files are now created with the `.yaml` extension (the official YAML extension).
- Both `.yaml` and `.yml` extensions are now detected and loaded.
- The config loader tries `.yaml` first, then falls back to `.yml` for backwards compatibility.
- Updated all CLI commands (`espanso edit`, `espanso path`) to use `.yaml` extension.
- Updated help text and error messages to reflect both extensions.
- Renamed resource files: `base.yml` → `base.yaml`, `default.yml` → `default.yaml`.

Existing users with `.yml` files will continue to work. The file watcher already supported both extensions, and the config loader now has a fallback mechanism.

Fixes #2184, #385.


AucaCoyan:
Thanks gcailly!
Hey @smeech this PR changes behaviour around the yaml extensions.

Currently we support .yml on the config only. It's a bug that we don't support .yaml extension. This is fixed with this PR here, but my question is if we should make the default, and first option to lookup YAML, or we stay with the current default YML. What do you think?

AucaCoyan:
Well, I made up my mind again, 😆 .
Let's merge!

AucaCoyan:
Smeech, please update the documentation, as this is important info for our users 🙏🏼 👍🏼

smeech:
> Smeech, please update the documentation, as this is important info for our users 🙏🏼 👍🏼

I'll add it to the list, but it won't affect most users, and none at all until it makes it into a Release.

If we're making it the default, however, and changing the names of `default.yml` and `config.yml` etc, that's a major change throughout the docs. **I would have set it as an _option_ for that reason alone!** There appear to be 130 occurrences of ".yml" in the docs. People on different versions reading the docs will need to be reminded this is from one particular version (that most are not using) onwards.

gcailly:
@smeech, I didn't think this change would impact the documentation so significantly. I'll make a PR for the docs.

smeech:
As I provide a lot of the support for new users, I try to think of how the docs appear for them. At this stage, we'll have many using the versions _before_ the change, as well as a handful on the new version. Bearing in mind we don't push updates, both need to understand what they're reading and how it affects their own configuration.

I don't want to have to have 130 instances of something like "if you're using v2.2.9 onwards, you can also use `default.yaml`" or whatever, and then have to remove them at a later date. I'm exaggerating a bit, but you get the point.

TBH, I would prefer to keep the existing defaults, but simply _permit_ `.yaml` files. It can be mentioned in a few places early in the documentation, for new users. Once the new version is widely established, we can switch the default files over and update the documentation as a whole. By then, users of older versions won't really be affected because they'll no longer be referring to the introductory sections.

Alternatives are to revert to versioned docs, as we did while v0.7.5 was prevalent, or introduce a "What's new" section we expect everyone to read through.

Without having had much time to think this through, I'm open to other suggestions, however!

gcailly:
@smeech I understand your point.
I made a PR anyway, in case you need it later. 🙂

AucaCoyan:
Hey!
You brought a really good point. I forgot again we document just for the stable version, and this is yet unreleased 😆 . About the topic, I think I passes for similar states of what you thought Smeech,

1) State: hey, this is great! should we merge?
In this position it was tempting to merge, and asked myself the consequences.
- It's a breaking change, but it has backwards compatibility
- many people use espanso since a long time ago, and persons get used to the software works over time. (Especially if there are few or non releases for a long time).
2) State: we should maintain the current functionality and priority order (first yml, then yaml):
If we change this order, this would mean anyone is happy and more users are welcome from the point in time this gets released (probably today).
But it kept bugging me: why do we use .yml in the first place? I thought at first that extensions need to have 3 letters, and it's just a convention. Well, it's not fully agreed, we have mpeg4, or .json or even longer extensions like .postman-collection. I read [your comment here](https://github.com/espanso/espanso/issues/2184#issuecomment-2560478163) and guess "well, smeech has already agreed upon this".
I browsed the web for responses and turns out that Windows didn't support 3 letter extensions waaaaaay back in the past (before windows 95!).
3) State: merge and move forward
At this point I thought that we aren't blocked by any substantial reason:
- the yaml extension is default onwards, giving the users a recognizable and easy extension, (we all say yaml in chat and threads but you write yml?)
- backwards users still get the compatibility if they are using .yml and upgrading to a newer version. You can mix and match even! This makes everyone happy.

I understand your pain to address possibly documenting multiple versions (and loads of work) and that not what we want ever again. Maintaining multiple versions of a documentation is a colossal burden to manage.
We have an application, not a library, we expect our users to update when a new version is released.
Whether the user doesn't know if there is a new version (1)  or the user is unable to update because newer versions doesn't run in his/hers OS, or simply are broken are all cases for a different discussion.

Maybe we (the espanso org) need to express our expectations we have over the users, so it's a clear understanding about what we provide, and back our comments and decisions with a public contract.

What do you think?

My suggestion for the documentation is NOT to change `.yml` for `.yaml` everywhere, but to mention it in a couple of places that it is allowed (maybe once, or twice).
If any user complains, then we can answer to update the software.
Smeech will have the final decision here 💯 

notes:
1- there is no notification of an update whatsoever. A self-update mechanism is a feature yet in the backlog

smeech:
> My suggestion for the documentation is NOT to change `.yml` for `.yaml` everywhere, but to mention it in a couple of places that it is allowed (maybe once, or twice).

My thoughts exactly.

AucaCoyan:
At the end I reverted the priority of `yaml` over `yml`, we had a chat with the team and decided that we should continue with the same priority.
In the future, we will pick yaml over yml! but we are not there yet 👍🏼

smeech:
Ideally we'd allow `.yaml` files in Espanso now.

> * Both `.yaml` and `.yml` extensions are now detected and loaded.

Priority doesn't really matter (except if users have duplicate base-named files with different extensions), but will probably cause fewer user issues if `.yml` loading is prioritised. Remember, Federico's preferences for quoted YAML values, and colon-prefixes, have been quite persistent!

> * The config loader tries `.yaml` first, then falls back to `.yml` for backwards compatibility.

No problem with:

> * Updated help text and error messages to reflect both extensions.

It's an excellent PR from @gcailly, it just goes a bit further than the Issues and what we need at the moment, I'm afraid :grin: .
Would it be possible to go ahead with these bits while they're fresh in our minds, and keep the rest for later?

AucaCoyan:
Yes! I didn't revert the whole PR, just changed the priority order from
1. yaml
2. yml

to
1. yml
2. yaml

In the future it is as easy as reverting the #2523 (reverting the revert 😆 ) and it's done

PR #2521, opened 2025-10-07
Prefer the GHA-provided rust tooling

n8henrie:
GitHub Actions can potentially be a significant security vulnerability;
we should try to avoid unnecessary dependencies.

The recent breakage seems to have been a transient issue; going
forward, I would advise that -- should we face such an issue again
-- we try temporarily pinning our runner image or our action to a
stable version (e.g. `ubuntu-24.04` as opposed to `ubuntu-latest`,
`actions/checkout@v4` as opposed to `actions/checkout@main`). Then we
can just revert back to latest or main once the issue has sorted itself
out.

### DIFF
--- a/.github/workflows/ci-arm.yml
+++ b/.github/workflows/ci-arm.yml
@@ -15,9 +15,6 @@ jobs:
     runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@main
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          components: clippy, rustfmt
       #   - name: Format
       #     run: cargo fmt --all -- --check
       - name: Install Linux X11 dependencies
@@ -57,9 +54,6 @@ jobs:
     runs-on: windows-11-arm
     steps:
       - uses: actions/checkout@main
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          components: clippy, rustfmt
       #   - name: Format
       #     run: cargo fmt --all -- --check
       - name: Build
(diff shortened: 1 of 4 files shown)


AucaCoyan:
Good! Nice move.
I added the x86 toolchain that was missing,

Upon a fast review, I realized we do not test windows or mac binaries. If we don't have specific functions to test in those OSes, that's fine. I have to check that

n8henrie:
> Upon a fast review, I realized we do not test windows or mac binaries.

Good catch. We probably should, though it will add some time to our CI builds.

We could probably recover some of that with caching though -- there is now an official GHA-provided cache that should work for rust builds. I tested it out at https://github.com/n8henrie/test-ga-fmt-cache/blob/master/.github/workflows/ci.yml

n8henrie:
Also, to be clear, I'm not 100% against depending on third-party actions, I just think it should be avoided. I am fine with depending on the official GitHub-provided actions, as I would expect them to be more secure (and quickly updated when vulnerabilities are discovered).

For example, I'm pretty sure that a bad actor with access to an action upon which we depend could access our secrets, potentially compromising your codesigning certificate, which would be a mess. (The secrets are masked in the output, but they might be able to e.g. start an interactive tmate session, or possibly echo those secrets to a file and then upload the file as an artifact, etc.).

n8henrie:
Well that failure is interesting.

```
error[E0463]: can't find crate for `core`
```

AucaCoyan:
I was because it was missing the macos x86 target.
Dunno if it's specified for arm but not for x86, and cargo pulls a different crate, etc.

I should merge, right?

n8henrie:
Oh, of course!

Yes, if it looks good to you, go ahead and merge.

I think I'm going to make a separate PR adding the cache.

PR #2522, opened 2025-10-08
use actions cache

n8henrie:
- **Cache rust artifacts with the official GHA actions/cache**
- **We want to run CI on pushes, not just PRs**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,13 @@ jobs:
   linux-x11:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/cache@main
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: actions/checkout@main
       - name: Format
         run: cargo fmt --all -- --check
@@ -72,6 +79,13 @@ jobs:
   linux-wayland:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/cache@main
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: actions/checkout@main
       - name: Format
         run: cargo fmt --all -- --check
@@ -110,6 +124,13 @@ jobs:
   macos:
     runs-on: macos-latest
     steps:
+      - uses: actions/cache@main
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: actions/checkout@main
       - name: Format
         run: cargo fmt --all -- --check
@@ -126,6 +147,13 @@ jobs:
   windows:
     runs-on: windows-latest
     steps:
+      - uses: actions/cache@main
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
       - uses: actions/checkout@main
       - name: Format
         run: cargo fmt --all -- --check
--- a/.github/workflows/ci-arm.yml
+++ b/.github/workflows/ci-arm.yml
@@ -14,6 +14,13 @@ jobs:
   linux-arm:
     runs-on: ubuntu-24.04-arm
     steps:
+      - uses: actions/cache@main
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  push:
```
AucaCoyan: This is not needed. `pull_request` re-runs the CI each time the PR gets updated (force push or normal push).
`push` is a bit broader. Includes the previous plus add tags, or any type of push event.
Also, leaving `push` here would trigger the CI workflow when a commit is put on top of the dev branch (the squash and rebase event when you close a PR). I don't want that, I want the `dev-release` run each time a commit is on the `dev`, and `ci.yml` work just on PRs.

Now that I describe the difference, it would make more sense that this file gets renamed to "pull_requests" instead of CI.
Every workflow is a CI, if you use it like this

AucaCoyan (changes requested):
Aside from this `push` event, all caches are great!

Thanks!

.github/workflows/ci.yml
n8henrie: Sounds good -- this is just a different workflow than I've become accustomed to in other multi-dev projects.

Usually, I am used to creating a development branch for a new feature or fix, iterating on it until things look good locally, then pushing this branch to the project. Once pushed, CI runs and ensures that my changes that were working locally also work in the shared environment, and also ensures that I haven't broken anything on other platforms (for example I do not have a windows system, my macos system is ARM, and all my Linux systems are on Wayland, so I can't really test for windows, Linux on X11, macos-x86, etc.).

Once I have verified that my changes are passing CI, and I've deemed that my changes are cleaned up and ready to be merged, only then do I make a PR to the development branch (`dev` in this case).

In my experience, the functionality is usually divided so that CI runs on every push (to every branch and PR), and the CD only runs on tags. This way it becomes super obvious what commit broke or fixed something, so we don't have to dive deeply into history and bisects when a PR fails CI (as the PR may contain many new commits, each of which could be the culprit).

Is your concern about running CI for every push mostly about exhausting the minutes for the free tier? Or perhaps something else entirely?

AucaCoyan: That sounds very reasonable. No, I'm not worried bout running out of the minutes, never hitted the limit, so I have no idea how much longer can it holds.

In the past on each merge commit a lot of workflows started: all the platforms of the dev-release, and again all the platforms for the CI job.
Given that the two of them do (mostly) the same, and dev-release does some packaging to have binaries, I decided to remove the CI job on dev-commits.

A couple of different approaches can be made to manage the CI, 
- some people use different files for each job (one file for format, one for lint, one for test, etc) and then they have "event files" that call upon different files, so you format when you commit a PR, but it's not needed for releasing the project (should be formated already). This is modular and nice, but it has a loooot of files
- other people do one file per event, and just copy and paste the steps for different jobs
- and I've seen another way, that they do just one file, and depending of the trigger they do different actions.
- and one more is putting everything in a Justfile and make the CI run the just file for different OSes

Additionally, you can set steps before and after another, so you can have a "block" of basics (formatting, compiling and fast tests) and then you start with the most expensive jobs (to test on every python version supported, with different dependencies, feature flags, etc, tokio-rs has one like this)

Any approach does the job well! If I had to pick one, I would prefer the one that gives the widest possibilities to devs to work (we welcome all), while at the same time being simple to read, modify and bugfix.

To take some action on this: What is your current need(s) that our CI doesn't cut it effectively?

n8henrie: Yes, lots of options! And I have to admit that my personal bias is usually to minimize complexity and minimize dependencies (we've touched on this briefly in the past). I've actually been quite happy with just a "CI" and "CD" workflow for most projects:
- CI runs on every commit and PR and ensures that for all target platforms a project compiles, tests pass, and lastly checks formatting and lints for best practices. In some projects it also packages a debug executable and uploads it as an artifact, which exercises the packaging steps and provides a handy way for end-users to help debug tricky to replicate issues.
- CD runs only for tags and does packaging, signing, and notarizing steps, uploads the final release to the website (if applicable), updates release notes.

As an aside, it's also *really* nice when the steps taken in CI are packaged into locally runnable scripts; for example, I can happily make an Espanso.app with a simple 'cargo build --release && scripts/create_bundle.sh' (I forget the actual script name, currently on mobile).

My concern with current state is that CI didn't run when I pushed this branch. I had to add the `push:` to get it to run CI so I would know whether or not I had broken anything.

The workflow I described above:

1. Develop on a work branch
2. Push to CI once passing locally
3. Make PR once CI passes and ready for merge

... works differently than what we have.

But I'm no expert, and I'm game to learn new processes!

What are your thoughts about managing more exploratory or experimental work? For example, if I were theoretically iterating on something that was not yet ready to be merged, and it just happened to break compilation on Windows. Wouldn't it be nice to know that sooner rather than later? With the current process, how would you approach this?

n8henrie: Will remove the `push:` to unblock this pending further discussion :)

AucaCoyan: Hello! Sorry I delayed so much the topic. It was an awesome conversation and I just left the phone on the table 😆 .
I'll review the CI topic and maybe we can create another issue to discuss it further

AucaCoyan (approved):

PR #2529, opened 2025-10-11
Add support for dynamic offset variables in date extension

gcailly:
Adds support for dynamic offset variables in the date extension.
The `offset` parameter can now accept both numbers and strings (for variable injection).
Maintains full backward compatibility with existing numeric offsets.


smeech:
Thank you!

smeech:
Now I just need an Espanso `eval` function instead of my shell scripts!

PR #2558, opened 2025-11-05
fix: check for both .yml and .yaml extensions before generating default files

### LINKED ISSUE #2557: base.yaml ignored when base.yml is auto-generated
### Description

Espanso auto-generates `base.yml` on startup even when `base.yaml` already exists, causing `.yaml` files to be ignored. This contradicts YAML.org's official `.yaml` recommendation and existing issues #385, #1316, #2184.

**Root cause:** 
https://github.com/espanso/espanso/blob/dev/espanso/src/config.rs#L57-L71

```rust
let match_file = sub_match_dir.join("base.yml");

if !match_file.is_file() {
    info!(
        "populating base.yml file with initial content: {}",
        match_file.display()
    );
    std::fs::write(match_file, DEFAULT_MATCH_FILE_CONTENT)?;
}
```

The code only checks for `base.yml`, not `base.yaml`.

### Steps to reproduce

1. Create `~/.config/espanso/match/base.yaml` with custom matches
2. Delete `base.yml` if it exists
3. Run `espanso start`
4. Observe: `base.yml` is auto-generated, `base.yaml` is ignored

### Expected behavior

Don't generate `base.yml` if `base.yaml` exists.

### Screenshots

_No response_

### Logs

Espanso starts normally but only loads auto-generated `base.yml`.

### Your environment

- Espanso version: 2.3.0
- OS: affects all platforms

### Self-service

- [x] I'd be willing to fix this bug myself.

mefengl:
Closes #2557

This PR fixes the issue where Espanso auto-generates `base.yml` even when `base.yaml` already exists, causing `.yaml` files to be ignored.

**Changes:**
- Check for both `.yml` and `.yaml` extensions before generating default config files
- Only generate if NEITHER extension exists

**Impact:**
- Minimal: only affects first-time setup or when config files are missing
- Backwards compatible: existing behavior unchanged

### DIFF
--- a/espanso/src/config.rs
+++ b/espanso/src/config.rs
@@ -54,16 +54,18 @@ pub fn populate_default_config(config_dir: &Path) -> Result<()> {
     }
 
     let default_file = sub_config_dir.join("default.yml");
+    let default_file_yaml = sub_config_dir.join("default.yaml");
     let match_file = sub_match_dir.join("base.yml");
+    let match_file_yaml = sub_match_dir.join("base.yaml");
 
-    if !default_file.is_file() {
+    if !default_file.is_file() && !default_file_yaml.is_file() {
         info!(
             "populating default.yml file with initial content: {}",
             default_file.display()
         );
         std::fs::write(default_file, DEFAULT_CONFIG_FILE_CONTENT)?;
     }
-    if !match_file.is_file() {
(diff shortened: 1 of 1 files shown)


AucaCoyan (approved):

AucaCoyan:
Thank you! I didn't catch this in the test phase 🚀

PR #2574, opened 2025-11-13
Restore espansion in the welcome screen

### LINKED ISSUE #2571: Espanso stopped expanding in the "Espanso is running!" welcome screen
### Description

Verified on macos and linux machines -- the most recent (2.3 and dev) versions of espanso no longer expand in the `Espanso is running!` welcome screen. Once that screen is closed normal function returns.

This was previously working fine in 2.2.? -- will try to bisect.

### Steps to reproduce

1. kill any running espanso processes
2. clean caches (and `Application Support/espanso` if on mac) so the welcome screen will pop up on next run
3. run `espanso launcher` or `open target/mac/Espanso.app --args launcher`
4. attempt to expand something in the `Espanso is running!` welcome screen

### Expected behavior

Expansion should work here! This will be users' first experience!

### Screenshots

_No response_

### Logs

_No response_

### Your environment

- Public source code: tested on current `dev` and `2.3.0`
- Operating system and version (e.g., Ubuntu 20.04.2 LTS): macos 26.1, nixos
- Espanso version (e.g., 2.1.4-beta): as per above

### Self-service

- [x] I'd be willing to fix this bug myself.

n8henrie:
The addition of the `wait` in 7bae4770073eac000c7711df8af8925509cefd1c broke expansion in the welcome screen.

Fixes https://github.com/espanso/espanso/issues/2571

### DIFF
--- a/espanso/src/cli/worker/engine/mod.rs
+++ b/espanso/src/cli/worker/engine/mod.rs
@@ -282,6 +282,7 @@ pub fn initialize_and_spawn(
                 }
             }
 
+            let mut welcome_handle = None;
             match start_reason.as_deref() {
                 Some(flag) if flag == WORKER_START_REASON_CONFIG_CHANGED => {
                     notification_manager.notify_config_reloaded(false);
@@ -296,7 +297,7 @@ pub fn initialize_and_spawn(
                     notification_manager.notify_start();
 
                     if !preferences.has_displayed_welcome() {
-                        super::ui::welcome::show_welcome_screen();
+                        welcome_handle = super::ui::welcome::show_welcome_screen();
                         preferences.set_has_displayed_welcome(true);
                     }
                 }
@@ -304,6 +305,9 @@ pub fn initialize_and_spawn(
 
             let mut engine = espanso_engine::Engine::new(&funnel, &mut processor, &dispatcher);
             let exit_mode = engine.run();
+            if let Some(mut handle) = welcome_handle {
+                handle.wait().expect("welcome screen died");
+            };
 
             info!("engine eventloop has terminated, propagating exit event...");
             ui_remote.exit();
--- a/espanso/src/cli/worker/ui/welcome.rs
+++ b/espanso/src/cli/worker/ui/welcome.rs
@@ -16,20 +16,18 @@
  * You should have received a copy of the GNU General Public License
  * along with espanso.  If not, see <https://www.gnu.org/licenses/>.
  */
+use std::process::Child;
 
 #[cfg(feature = "modulo")]
-pub fn show_welcome_screen() {
+pub fn show_welcome_screen() -> Option<Child> {
     let espanso_exe_path = std::env::current_exe().expect("unable to determine executable path");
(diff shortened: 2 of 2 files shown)


n8henrie:
@AucaCoyan this was the issue -- after 7bae4770073eac000c7711df8af8925509cefd1c espanso `wait`s for the welcome screen to complete in this thread, so nothing else works until the welcome screen is dismissed.

Removing the `wait` here allows this thread to spawn and everything else to continue in the background.

It might be advisable for us to `join` the thread somewhere? Although I presume it will just go away once the welcome screen is dismissed.

Before this is merged, we should probably:
- [ ] decide if we need to join this thread somewhere
- [ ] A similar change (`wait`) was added to [espanso-engine/src/process/middleware/open_config.rs](https://github.com/espanso/espanso/commit/7bae4770073eac000c7711df8af8925509cefd1c#diff-bbf8c89e010981d4b57eafe80fec266b3a8ab893714e4f47dd2c45c3bdfa5adbL68) in this commit -- did this introduce any undesirable behavior? I usually use immutable configs (via home-manager) and therefore don't use the config editing and would not have noticed. Should investigate.

AucaCoyan:
Interestingly, clippy warns of the process not being awaited:
```console
error: spawned process is never `wait()`ed on
  --> espanso/src/cli/worker/ui/welcome.rs:26:13
   |
26 |     let _ = command.spawn().expect("unable to show welcome screen");
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^- help: try: `.wait()`
   |
   = note: not doing so might leave behind zombie processes
   = note: see https://doc.rust-lang.org/stable/std/process/struct.Child.html#warning
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#zombie_processes
   = note: `-D clippy::zombie-processes` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::zombie_processes)]`

error: could not compile `espanso` (bin "espanso") due to 1 previous error
Error: Process completed with exit code 101.
```

this was the reason that commit had a `wait`, I apply all the clippy fixes no matter what does it mean the contents.
Usually this means no behaviour change, but you found an interesting bug 🤔 


I'll review this later! Thanks for taking hands into the keyboard!

n8henrie:
@AucaCoyan the latest push adds a wait, just after the worker threads are spawned, and seems to work locally for me.

Let me know what you think -- if we can get this in shape to merge, I'm hoping to get a meaningful espanso update (esp for wayland) merged into nixpkgs for the 25.11 release (cutoff coming up soon).

n8henrie:
Had to fix compilation on non-modulo builds but think I got it now.

AucaCoyan:
oh, non-modulo builds are not in my radar of problems!
It is a nice to have feature, but I think you would broke half of espanso if you try to use it without modulo. It may work for simple triggers, but I don't know what ramifications could happen if you try a form, or open the search bar, does it panics and dies? dunno, and currently not a worry to be honest

AucaCoyan:
Excellent! I liked the solution very much!

PR #2590, opened 2025-12-02
Docs update: Improving build notes around Windows and WSL

ad08fee3:
I'm a new contributor and also a Rust noob, and I spent a loooong time trying to figure out how to get the build to work.

For my job I mostly do my dev work on Ubuntu, but at home I'm chained to Windows. As a result, I try to do most of my at-home dev work in WSL instead of Windows natively. I thought I would be able to do that with Espanso, but I spent a long time trying to get it to work. I tried a bunch of workarounds and did a lot of learning about Rust cross-compilation, but in the end I ended up deciding it wasn't worth the massive code changes it would require. So I bailed on my cross-compilation goals, sadly.

However, I am a weirdo who enjoys writing code documentation so I'm documenting my findings about the build on Windows and cleaning up some other ambiguity I found.

Hopefully these changes will help save the time of future contributors and make it easier for new folks to learn how stuff works. 😀

### DIFF
--- a/docs/src/ch03-00-building-and-compilation.md
+++ b/docs/src/ch03-00-building-and-compilation.md
@@ -2,6 +2,10 @@
 
 This chapter explains the various steps needed to build espanso.
 
+### Cross-compilation
+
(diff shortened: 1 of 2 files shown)


AucaCoyan:
Thank you! These contribs are more than welcome!

Issue #2582, opened 2025-11-21
espanso only working in every other window on sway


n8henrie:
### Description

This one is super bizarre... but on sway, on two different systems (arch and nixos), Espanso is working fine only in *every other window* that I bring up.' See videos below in the screenshots section.

### Steps to reproduce

1. Open a window. Try to expand a snippet.
2. Open another window. Try again.

### Expected behavior

Proper expansion in every window

### Screenshots

In alacritty:

https://github.com/user-attachments/assets/16cec828-5ae2-48b7-803f-ead99f6a1d4b

Same behavior in anyrun:

https://github.com/user-attachments/assets/45bdae1b-e442-45af-97dc-d6bfb512bf04

### Logs

Nothing of interest. No output under default log level when expanding in one window and not expanding in the next.

### Your environment

Same bug both on nixos (nixos-unstable) using espanso-wayland from nixpkgs as well as arch linux compiled from source.

- Environment name and version (e.g., Rust 1.59.0): Rust 1.91.0
- Operating system and version (e.g., Ubuntu 20.04.2 LTS): As per above
- Espanso version (e.g., 2.1.4-beta): 2.3.0

### Self-service

- [x] I'd be willing to fix this bug myself.

n8henrie:
Same behavior present back to at least 2.2.1

Issue #2497, opened 2025-10-01
Wayland and a ch_FR keyboard layout results in weird behaviour


iGormilhit:
### Description

The last release of the Gnome desktop environment sets Wayland by default, without the possibility to switch to Xorg, at least on Arch Linux. So I've adapted my configuration, accordingly to the documentation.

Since the last Espanso (the espanso-wayland package from the arch user repository), I encounter weird behaviours.

First, I had to set the default backend to Inject. In deed, with the auto setting, when it uses the Clipboard backend, espanso just add the last entry of the clipboard instead of the content of the match.

But, with the Inject backend, I have two issues (which may be the same):

1. I have matches that contains a narrow non-breakable space (as it is used in some French typos). When I trigger them, nothing happens and in the espanso log, I get this message: "missing vkey mapping for char ' '".
2. I have a math to type "Ç", and when I trigger it, I usually get the "Ç", but then the capslock key is activated. I then have to deactivate the capslock key and to restart espanso. But no error message in the espanso log.

My espanso configuration:

```
backend: Inject

keyboard_layout:
  layout: "ch"
  variant: "fr"

preserve_clipboard: true
show_notifications: false
```

Some of my matches:

```
  # Espace insécable fine
  - trigger: ";f"
    replace: " "

  # C cédille majuscule
  - trigger: ";C"
    replace: "Ç"
```

### Steps to reproduce

1. With Gnome using Wayland.
2. Install `espanso-wayland` from AUR.
3. Use the configuration provided before.
4. Set your keyboard to the ch_FR layout (Swiss French).
5. Use the match rules provided before.
6. Try to trigger the two matches.

### Expected behavior

Expand the content of the replace, and do not modify the state of the keyboard.

### Logs

```
11:28:06 [daemon(89520)] [INFO] reading configs from: "/home/igor/.config/espanso"
11:28:06 [daemon(89520)] [INFO] reading packages from: "/home/igor/.config/espanso/match/packages"
11:28:06 [daemon(89520)] [INFO] using runtime dir: "/home/igor/.cache/espanso"
11:28:06 [daemon(89520)] [INFO] system info: Arch Linux v - kernel: 6.16.8-arch3-1
11:28:06 [daemon(89520)] [INFO] kdotool missing or not available for the current wayland DE.
11:28:06 [daemon(89520)] [INFO] espanso version: 2.2.7
11:28:06 [daemon(89520)] [INFO] spawning the worker process...
[...]

Issue #2428, opened 2025-08-06
In Ghostty terminal under Wayland both backends do not work


kekscode:
### Description

Instead of the completion, the most recent item from the clipboard is pasted every time. In other applications, like my Chrome-based Webbrowser, Electron-based stuff and KDE applications like Konsole, the configured completions work without issues.

Note: The same configuration also works on MacOS in Ghostty, if this matters.

The espanso daemon config is:

```yaml
search_shortcut: ""
auto_restart: true
backend: auto
```

I tried all three possible values for `backend`. The `auto` value seems to select `Clipboard` i think.

### Steps to reproduce

1. Open Ghostty
2. Execute any completion

### Expected behavior

Completion should be written, but last item from clipboard is shown for any completion i tried.

### Your environment

- Public source code: ?
- Environment name and version (e.g., Rust 1.59.0): rustc 1.88.0 (6b00bc388 2025-06-23) (Arch Linux rust 1:1.88.0-1.1)
- Operating system and version (e.g., Ubuntu 20.04.2 LTS): CachyOS Linux (Arch-based)
- Espanso version (e.g., 2.1.4-beta): 2.2.4
- ghostty version=1.1.3-arch1.1
- Wayland (no x11)

### Self-service

- [x] I'd be willing to fix this bug myself (i don't know rust, though)

smeech:
Is `restore_clipboard_delay` from [options](https://espanso.org/docs/configuration/options/#options-reference) any help?

On Wayland, this is frequently about specifying the `keyboard_layout:` parameters correctly, however.

kekscode:
I tried various `restore_clipboard_delay` values from `50` to `10000`, but no success.

Checking `keyboard_layout` now. The log actually says:

```
23:41:19 [worker(13377)] [WARN] unable to determine keyboard layout automat
ically, please explicitly specify it in the configuration.
[...]
23:41:19 [worker(13377)] [INFO] using WaylandFallbackClipboard

```

kekscode:
I checked `/etc/default/keyboard` and added this snippet (and also did some try/error changes):

```yaml
keyboard_layout:
  layout: "us"
  model: "pc105"
```

no change at all, still working everywhere i try except for ghostty.

For the sake of completeness, the logs reflected the change:

```
23:54:30 [worker(19587)] [INFO] inject module will use this keyboard layout
: [R=, M=pc105, L=us, V=, O=]
[...]
[...]

Issue #2340, opened 2025-05-17
It seems, that on wayland neither the clipboard nor typing work.


alfonz19:
### Description

Hi, it seems, that on wayland neither the clipboard nor typing work.

Having configuration:
```
matches:
  - trigger: "::tel"
    replace: "123456789"
```
and NON-US keyboard layout on WAYLAND, will not produce numbers, but chars, which my current layout lies on keys 123456789. 

In attempt to fix it, I lowered clipboard threshold to 1, but this does not work either, because espanso does not put the 123456789 data into clipboard and paste what's already in clipboard instead. (I have wl-clipboard packae installed, and wl-copy and wl-paste commands are available) 

How to fix?

(ubuntu 24.10, wayland)

### Steps to reproduce

I guess install on fresh wayland distribution?

### Expected behavior

for trigger ::tel as shown above, I'd expect 123456789 printed regardless of current keyboard layout, especially if clipboard threshold is set to 1.

### Your environment

- Public source code:
- Environment name and version (e.g., Rust 1.59.0):
- Operating system and version (e.g., Ubuntu 24.10):
- Espanso version (e.g., 2.2.1):

### Self-service

- [ ] I'd be willing to fix this bug myself.  --> no idea how to debug.

smeech:
You may need to tweak your [`keyboard_layout`](https://espanso.org/docs/configuration/options/#options-reference) parameters. See #2338 for example.

alfonz19:
> You may need to tweak your [`keyboard_layout`](https://espanso.org/docs/configuration/options/#options-reference) parameters. See [#2338](https://github.com/espanso/espanso/issues/2338) for example.

Thank you for reply. 

So I added:

```
keyboard_layout:
  layout: "cz"
  variant: "alfonz"
```

to default.yml and updated threshold in other direction to prevent clipboard usage: 

`clipboard_threshold: 1000`


This help a little with US symbols to type, so for example with the one from provided example. But for other cases not so much. Lets add more examples then ;) :

```
matches:
  - trigger: "::dd"
    replace: "Dobrý den, "
  - trigger: "::mm"
    replace: "Martin Mucha"
  - trigger: "::tel"
    replace: "123456789"
  - trigger: "::uuid"
    replace: "{{output}}"
    vars:
      - name: output
        type: shell
        params:
          cmd: "uuidgen"
```

[...]

PR #2612, opened 2026-01-01
fix(wayland): copied text not paste-able in some programs

bew:
Hello!

(note: this was fully Human-made, zero AI involved)

> [!NOTE]
> Edit: Skimming GH issues mentioning `Wayland`, I think this PR could fix the following issues:
> #2428 #2162 #2270 #2497 #2582 #1670 #2340 #901 #1367
> (maybe more, to be confirmed)

### Context

Since I had to switch to Wayland a few months ago, I've always had trouble with Espanso and attributed it to some weirdness around Wayland, protocol instabilities, KDE Plasma6.. (and the missing capability on NixOS)

For example some expansions would work in Firefox but not all, and almost no expansions would work in Wezterm (the trigger is removed but the replacement is not pasted).
Other times it would simply paste the wrong clipboard data, as if Espanso didn't set the clipboard at all.
And when I had something like `foo` in my clipboard, tried an expansion that failed, I couldn't paste the original `foo` that was supposed to be restored by Espanso after the expansion.

Since latest NixOS 25.11 release, the capability issue was fully resolved and I was still having these weird issues..
The first day of 2026 motivated me to dive in and actually fix this! 😀

### Findings

With `WAYLAND_DEBUG=1` and some additional debug logs I've found that the MIME type of the copied text is set correctly then immediately overwritten to a bad type:
```
...
[ 442709.543] {Default Queue}  -> wl_data_device_manager#7.get_data_device(new id wl_data_device#3, wl_seat#6)
[ 442741.732] {Default Queue}  -> wl_data_device_manager#7.create_data_source(new id wl_data_source#11)
...

!! NOTE: The clipboard is set here
[ 442747.844] {Default Queue}  -> wl_data_device#3.set_selection(wl_data_source#11, 35716)

!! NOTE: The data type is set correctly here (`offer` line)
[ 442747.880] {Default Queue} wl_data_device#3.data_offer(new id wl_data_offer#4278190080)
[ 442747.891] {Default Queue} wl_data_offer#4278190080.offer("text/plain;charset=utf-8")   // <- here
[ 442747.900] {Default Queue} wl_data_device#3.selection(wl_data_offer#4278190080)
...

!! BAD: The data type is ~immediately overwritten to a bad type
[ 442748.373] {Default Queue} wl_data_device#3.data_offer(new id wl_data_offer#4278190081)
[ 442748.383] {Default Queue} wl_data_offer#4278190081.offer("l�4��U")   // <- here
[ 442748.390] {Default Queue} wl_data_device#3.selection(wl_data_offer#4278190081)
...
```
And this weird MIME type was also used when restoring the original clipboard content, easy to validate with `wl-paste --list-types`.

In Wezterm, I've found that only selection with MIME type `text/plain;charset=utf-8` can be pasted, so the weird `l�4��U` type doesn't cut it and nothing can be pasted after Espanso changed the clipboard.

It probably worked better in Firefox because the browser have a lot more flexibility on the kind of data that can be pasted, potentially doing its own detection.

### The fix

I'm not exactly sure why the MIME type was immediately overwritten after being correct for less than a second.
I tried to disable Klipper (KDE's clipboard manager) which might have caused this but I still had the same issue.
👉 Anyway, manually setting the MIME type fixes everything for me, and all my expansions work everywhere.

### Thoughts

The clipboard restoration might need a better impl, to actually restore the original MIME type instead of forcing plain text. I'd prefer to do this in a separate PR though.

Also is there a reason the `wl-clipboard` crate isn't used, instead of calling `wl-copy` & `wl-paste` cli?
(using the cli is probably easier to do / think about 👀)

### DIFF
--- a/espanso-clipboard/src/wayland/fallback/mod.rs
+++ b/espanso-clipboard/src/wayland/fallback/mod.rs
@@ -119,7 +119,15 @@ impl Clipboard for WaylandFallbackClipboard {
     }
 
     fn set_text(&self, text: &str, _: &ClipboardOperationOptions) -> anyhow::Result<()> {
-        self.invoke_command_with_timeout(&mut Command::new("wl-copy"), text.as_bytes(), "wl-copy")
+        // NOTE: Without explicit MIME type, wl-copy's auto-detection can give unexpected results
+        // making the text not paste-able in some programs.
+        self.invoke_command_with_timeout(
+            Command::new("wl-copy")
+                .arg("--type")
+                .arg("text/plain;charset=utf-8"),
+            text.as_bytes(),
+            "wl-copy",
+        )
     }
 
     fn set_image(
--- a/espanso-clipboard/src/wayland/fallback/mod.rs
+++ b/espanso-clipboard/src/wayland/fallback/mod.rs
@@ -77,45 +77,13 @@ impl WaylandFallbackClipboard {
 
 impl Clipboard for WaylandFallbackClipboard {
     fn get_text(&self, _: &ClipboardOperationOptions) -> Option<String> {
-        let timeout = std::time::Duration::from_millis(self.command_timeout);
-        match Command::new("wl-paste")
-            .arg("--no-newline")
-            .stdout(Stdio::piped())
-            .spawn()
-        {
-            Ok(mut child) => match child.wait_timeout(timeout) {
-                Ok(status_code) => {
-                    if let Some(status) = status_code {
-                        if status.success() {
-                            if let Some(mut io) = child.stdout {
-                                let mut output = Vec::new();
-                                io.read_to_end(&mut output).ok()?;
-                                Some(String::from_utf8_lossy(&output).to_string())
-                            } else {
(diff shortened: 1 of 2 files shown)


bew:
The build failures in CI are due to unused imports in files unrelated to this PR 👀

TaQuangKhoi:
I hope this PR will be reviewed and merged..

bew:
NOTE: It's currently an incomplete impl: when I have non-text (e.g. an image) in my clipboard before an expansion, I can't paste that non-text after the expansion.

Apart from that it's working very well on my NixOS system, on KDE Wayland 6.6 🚀 

For the record: I also tried to swap the implementation in https://github.com/bew/espanso/tree/impl-native-wayland to use the `wl-clipboard` crate (instead of the `wl-copy`/`wl-paste` binaries) but I've hit many crashes or times where it just didn't paste after removing the trigger, and only pasted later when I tried to paste something else. I didn't investigate much further at the moment.

PR #2618, opened 2026-01-12
feat: Add match_defaults support for Issue #2185

### LINKED ISSUE #2185: Default options for matches
### Description

There could/should be a way to have shared options (per-file or globally maybe) that are shared across all matches. e.g.
```yaml
matches:
  - trigger: "a"
    replace: "something"
    propagate_case: true
    word: true
  - trigger: "b"
    replace: "something else"
    propagate_case: true
    word: true
```
could be simplified to:
```yaml
global_options:
  propagate_case: true
  word: true
matches:
  - trigger: "a"
    replace: "something"
  - trigger: "b"
    replace: "something else"
```

Possible other names I can think of instead of `global_options` are `global_match_options` and `shared_match_options`.
Or maybe do something like this?
```yaml
matches:
  defaults:
    propagate_case: true
    word: true
```
But I don't think that would really work? I'm not super familiar with how Yaml is parsed...

### Motivation

Writing a bunch of matches with a lot of the same options can get really messy/verbose/complex.

### Alternatives

_No response_

- [ ] I'd be willing to contribute this feature to Espanso myself.

sboldt1978:
## Summary

This PR implements the `match_defaults` feature requested in #2185, allowing users to set default match options at the file level to reduce YAML repetition.

Resolves #2185

## What's New

Added support for a `match_defaults` section in YAML configuration files that provides default values for match behavior options:

- `word` - Default word boundary matching (both left and right)
- `left_word` - Default left word boundary
- `right_word` - Default right word boundary
- `propagate_case` - Default case propagation behavior
- `uppercase_style` - Default uppercase transformation style
- `force_clipboard` - Default clipboard forcing
- `force_mode` - Default injection mode (clipboard/keys)

## Usage Example

```yaml
match_defaults:
  word: true
  propagate_case: true
  uppercase_style: "capitalize_words"

matches:
  - trigger: ":address"
    replace: "123 main street"
    # Inherits: word=true, propagate_case=true, uppercase_style="capitalize_words"
  
  - trigger: ":inline"
    replace: "inline text"
    word: false
    # Overrides: word=false
    # Inherits: propagate_case=true, uppercase_style="capitalize_words"
  
  - trigger: ":code"
    replace: "const x = 42;"
    propagate_case: false
    # Overrides: propagate_case=false
    # Note: uppercase_style has no effect without propagate_case
```

## Precedence

The implementation follows this precedence chain (highest to lowest):
1. Individual match option
2. `match_defaults` option
3. Built-in espanso defaults

## Testing

- Added 11 comprehensive tests covering:
  - YAML parsing with match_defaults (5 tests)
  - Defaults inheritance and overrides (5 tests)
  - Edge cases: force_mode precedence, warning logic, regex matches (3 tests)
  - Case propagation verification (1 test)
- All 93 tests passing (79 existing + 14 new)
- Manually tested with live espanso instance on macOS
- Code formatted with `cargo fmt --all`
- Linted with `cargo clippy --workspace` (no warnings in new code)

## Implementation Details

**Modified files:**
- `espanso-config/src/matches/group/loader/yaml/parse.rs` - Added `YAMLMatchDefaults` struct
- `espanso-config/src/matches/group/loader/yaml/mod.rs` - Implemented precedence logic and defaults application

**Key technical decisions:**
- Used Rust's `Option` chaining (`.or()`) for clean precedence resolution
- Fixed edge case where `force_mode` precedence was incorrect
- Improved warning logic for `uppercase_style` without `propagate_case`
- Maintains full backward compatibility - existing configs work unchanged


smeech:
That looks impressive, and required (albeit quite extensive) changes to only two files to solve a five-year-old Issue!

I would point out that `force_clipboard:` is considered deprecated, and _could_ be omitted. I suspect `force_clipboard: false` has _no_ effect anyway, although I note the tests for it in `parse.rs`. `force_mode:` should work ideally, I suppose, although it's rather less important than the `word:` and `case` options, because it can be set in `default.yml` or app-specific configuration `backend:` values for the most part. Apart from testing recipient programs for the optimal injection method, its only use, IMO,  is for catering to different programs in Wayland, when app-specific configurations don't work.

Beyond that I'm not really qualified to comment as I'm not a coder, but I do have permission to run the automated checks, which is showing one problem across all platforms ☹️.

sboldt1978:
Thanks for running the automated tests. Even though I modified only two files, the CI jobs complained problems in several files which I hadn't modified at all. All of the errors where due to "use log::error;"  imports that are not used inside the file. 

The CI checks for pull request seem to be really strict to rate this as an "error" which is a show stopper for any code additions via PRs (at least for those which don't fix the existing problems in the first place).  However, I had to add an additional commit a98a55236327b0a4e5f8edb8e3a7e0eb9d1cb6e0 which removed the unnecessary  "use log::error;" entries that the CI complained about to be able to get a "green" result for the real changes. Now it passes all the checks for all OS. 
Kind request: You should apply the same strictness for all types of commits, because otherwise we might again end up again with a code level for which the CI checks report "false positives" for code wich wasn't changed by the contributor at all, as it was in my case.

PS: Yes, this issue / feature request is five years old, but I just switched from aText to espanso and this was one of the features I missed most. Therefore I implemented it mostly for my own needs. You (smeech) sent me the link to this long time open request on discord which is describing exactly what I need, so I linked my code to this issue / feature request :)  

It would be great if my PR would get merged soon, because I'm already working on another feature that depends on this PR, therefore I don't want to open a new PR for it without the dependency already being part of the dev branch.

PR #2654, opened 2026-03-14
fix(wayland): keep wl-copy alive to maintain clipboard ownership

DzmingLi:
## Problem

On Wayland, `WaylandFallbackClipboard::set_text` (and `set_html`/`set_image`) never actually sets the clipboard, causing paste operations to fall back to the previous clipboard owner. Users see old clipboard content inserted instead of the configured replacement text.

**Root cause:** `invoke_command_with_timeout` uses `child.stdin.as_mut()` to write data to `wl-copy`, which borrows stdin without closing the pipe. `wl-copy` blocks indefinitely waiting for EOF on stdin and never reads the data. After `wayland_command_timeout_ms` (default 2s), espanso kills the process — which has not set the clipboard at all. `set_text` then returns `Err`, and `inject_text` propagates it via `?`, so `send_paste_combination` is never called. Only the backspace deletion of the trigger occurs, with nothing replacing it (or the application pastes the previous clipboard content through its own means).

**Secondary issue:** Even if stdin were closed correctly, `wl-copy` must remain running after reading its input to maintain Wayland clipboard ownership and serve data requests from applications that paste. Killing it before the paste occurs causes the same fallback behaviour.

This affects all Wayland users using `backend: clipboard` and also `backend: auto` on Linux, which automatically routes non-ASCII text (e.g. Unicode symbols) through the clipboard injector (see `text_inject.rs`).

## Fix

- Replace `child.stdin.as_mut()` with `child.stdin.take()` so the write end of the pipe is closed (dropped) immediately after writing, sending EOF to `wl-copy` and allowing it to finish reading and enter its Wayland event loop.
- Rename `invoke_command_with_timeout` to `spawn_copy_command` and change the semantics: instead of waiting for the process to exit and killing it on timeout, do a short startup check (200ms) to detect immediate crashes, then store the running process in `active_copy_process` (a `Mutex<Option<Child>>` field) to keep it alive.
- Kill the previous `active_copy_process` when a new clipboard write occurs, and clean up in `Drop`.

## Relation to #2612

#2612 addresses a related but distinct symptom: it fixes the MIME type by explicitly passing `--type text/plain;charset=utf-8` to `wl-copy`, preventing clipboard managers from overwriting the selection with a garbled type.

This PR fixes the more fundamental issue upstream of that: because `child.stdin.as_mut()` never closes the pipe, `wl-copy` never receives EOF, never reads the data, and is killed before it sets the clipboard at all. At that point the MIME type is irrelevant — nothing was written. Additionally, even with a corrected stdin, killing `wl-copy` immediately after writing loses Wayland clipboard ownership before the paste occurs.

The two fixes are complementary and both may be worth merging.

### DIFF
--- a/espanso-clipboard/src/wayland/fallback/mod.rs
+++ b/espanso-clipboard/src/wayland/fallback/mod.rs
@@ -21,7 +21,8 @@ use std::{
     io::{Read, Write},
     os::unix::net::UnixStream,
     path::PathBuf,
-    process::Stdio,
+    process::{Child, Stdio},
+    sync::Mutex,
 };
 
 use crate::{Clipboard, ClipboardOperationOptions, ClipboardOptions};
@@ -33,6 +34,9 @@ use wait_timeout::ChildExt;
 
 pub(crate) struct WaylandFallbackClipboard {
     command_timeout: u64,
+    // Keep the wl-copy process alive to maintain Wayland clipboard ownership.
+    // The clipboard owner must remain running to serve paste requests.
+    active_copy_process: Mutex<Option<Child>>,
 }
 
 impl WaylandFallbackClipboard {
@@ -71,6 +75,7 @@ impl WaylandFallbackClipboard {
 
         Ok(Self {
             command_timeout: options.wayland_command_timeout_ms,
+            active_copy_process: Mutex::new(None),
         })
     }
 }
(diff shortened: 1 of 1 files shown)


bew:
Hello @DzmingLi, I was revisiting my old #2612 PR and wondering:
Was this the wrong impl to fix the issue you had?
Why did you close the PR ?

AucaCoyan:
Hey! I'm not sure, but it might be related that I'm struggling to review all the PRs and he decided to close it as he is no longer interested.
I didn't checked if it is already solved yet, nor read the changes

PR #2659, opened 2026-03-25
feat: add regex_transform variable extension for advanced text replacement

Zettt:
# Pull Request: Add regex_transform variable extension for advanced text replacement

## Summary

This PR introduces a new variable extension called `regex_transform` to the `espanso-render`.

The `regex_transform` extension allows users to perform regex-based find-and-replace operations directly on variable values (such as text captured by regex triggers) during text expansion. Additionally, it introduces a `modifiers` parameter, which allows users to apply case alterations (such as `lowercase`, `uppercase`, and `capitalize`) to specific captured groups.

## Motivation

A major use-case for this feature is recreating advanced auto-correct functionality found in tools like Typinator. For example, users often want to correct accidental double-caps at the beginning of words (e.g., `BAnk` -> `Bank`).

Previously, manipulating regex capture groups in Espanso was only possible using major workarounds, such as relying on external shell scripts. These workarounds are not only slow to execute during text expansion but also impractical to maintain across different operating systems. Solving this natively within Espanso using internal Rust functionality makes these text transformations significantly faster and much easier for users to configure directly in their YAML files.

### Example configuration

```yaml
matches:
    - regex: "\\b(?P<word>[A-Z]{2}[a-z]+)(?P<end>\\W)"
      replace: "{{corrected_word}}{{end}}"
      vars:
          - name: corrected_word
            type: regex_transform
            params:
                source: "{{word}}"
                find: "^([A-Z])([A-Z])(.*)$"
                replace: "$1$2$3"
                modifiers:
                    "2": "lowercase"
```

## Checklist

_Following the guidelines in `ch04-01-creating-a-pr.md`:_

- [x] Kept PR focused on a single functional change
- [x] Tested the changes locally on my machine
- [x] Added unit testing within the `tests` module in `regex_transform.rs`
- [x] Code is formatted via `cargo fmt --all`
- [x] Passes `cargo clippy --workspace`
- [x] All tests pass via `cargo test --workspace`
- [x] Added documentation for `regex_transform` (Please let me know where you'd prefer documentation for this to live!)

### DIFF
--- a/espanso-render/src/extension/mod.rs
+++ b/espanso-render/src/extension/mod.rs
@@ -26,4 +26,5 @@ pub mod form;
 pub mod random;
 pub mod script;
 pub mod shell;
+pub mod regex_transform;
 mod util;
--- a/espanso-render/src/extension/mod.rs
+++ b/espanso-render/src/extension/mod.rs
@@ -24,7 +24,7 @@ pub mod echo;
 mod exec_util;
 pub mod form;
 pub mod random;
+pub mod regex_transform;
 pub mod script;
 pub mod shell;
-pub mod regex_transform;
 mod util;
--- a/espanso-render/src/extension/regex_transform.rs
+++ b/espanso-render/src/extension/regex_transform.rs
@@ -102,7 +102,7 @@ impl Extension for RegexTransformExtension {
             // Because replace_all closure doesn't have a direct way to build the replacement string
             // with mapped capture groups while easily substituting them into the `replace` template,
             // we will expand the `replace` template manually using the captured groups.
-            
+
             // To do this simply, we parse the `replace` string for $N or ${N} and substitute them.
             let mut expanded = String::new();
             let mut chars = replace.chars().peekable();
@@ -135,7 +135,7 @@ impl Extension for RegexTransformExtension {
                     if let Ok(group_idx) = group_idx_str.parse::<usize>() {
                         if let Some(matched_group) = caps.get(group_idx) {
                             let mut group_val = matched_group.as_str().to_string();
-                            
+
                             // Apply modifier if configured
                             if let Some(modifier) = modifiers.get(&group_idx) {
                                 match modifier.as_str() {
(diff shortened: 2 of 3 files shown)


smeech:
I would value the intoduction of `modifiers:` to transform to upper- or lower-case characters, and perhaps other transformations, and we would really benefit from other conditional processes in Espanso that don't rely on external shell scripts, but does `regex_transform` actually add anything? I'm very familiar with the double-caps issue, and have a few [routines](https://gist.github.com/smeech/affa291af51c5bea7a8c1f6d589c0584#file-caps-yml) of my own (where you'll also see my need to have exceptions).

Can someone give me another example of the use of `regex_transform` which doesn't simply duplicate the process of the `regex:` trigger itself? In the above example one might use something like:
```yml
  - regex: (?P<char1>\b[[:upper:]])(?P<char2>[[:upper:]])(?P<end>[[:lower:]]{2,}\W)
```
performing a transformation on `{{char2}}` before returning the string `{{char1}}{{char2}}{{end}}` without the intermediate step. Alternatively, one would capture the whole word and perform a `capitalise` or `capitalize` (😀) transformation to it anyway.

I don't want to be critical, but am trying to avoid duplication.

Zettt:
Thank you for the review, @smeech.

Direct case transformations were on my personal wishlist for this feature. I held off to keep this pull request focused and conform to contribution guidelines. I am open to adding direct transform options if demand exists.

`regex_transform` and the trigger regex serve different roles:

1. Detection versus mutation: Triggers capture keystrokes. They cannot alter casing or replace patterns. `regex_transform` mutates the text.
2. Non-trigger data: `regex_transform` modifies variables like `{{clipboard}}` or form fields. Triggers cannot inspect clipboard contents or clean copied links.
3. Iterative substitution: `regex_transform` runs `replace_all`. Triggers perform a single match and cannot replace multiple pattern instances, such as converting spaces to hyphens.

Your gist illustrates the issue with script workarounds.Spawning Python, Bash, or PowerShell is just too slow. I figured executing regex directly inside the compiled Rust binary removes those delays and functions the same across multiple platforms.

smeech:
> 2. Non-trigger data: `regex_transform` modifies variables like `{{clipboard}}` or form fields. Triggers cannot inspect clipboard contents or clean copied links.
> 3. Iterative substitution: `regex_transform` runs `replace_all`. Triggers perform a single match and cannot replace multiple pattern instances, such as converting spaces to hyphens.

:+1:  Thank you - that helps a lot.

I'm very please to see the modifier feature - it's an obvious addition for a text expander and has been on my list for a long time (in addition to arithmetic, and `if … then` constructs which your addition may help workaround for some cases).

PR #2671, opened 2026-04-13
feat: add label/id support for choice fields in forms #2627

### LINKED ISSUE #2627: Choice Forms with labels/ids don't work on a form with a text box
### Description

From what I can tell, a "choice" form just doesn't drop down to show the choices if it's present in a form with another type of field (like a text field). I tried the snippet outlined below on both my mac (macOS Tahoe 26.2) and my Linux machine running Pop!_OS 24.04/Kernel 6.17.9-76061709.

I'm curious if this is because there's not currently a way to render the label/id picker in-line the same way you can with the normal choice picker? 

### Steps to reproduce

1. Add the following to your base.yml:
```
  - trigger: ";bug "
    form: |
      This is the prompt UI. Enter text here:
      [[text_box]]
      Here is a choice picker:
      [[choice_picker]]
    form_fields:
      choice_picker:
        type: choice
        values:
          - label: "Label 1"
            id: "ID 1"
          - label: "Label 2"
            id: "ID 2"
```
2. Use the trigger to expand it
3. RESULT: The choice drop down is blank and clicking it doesn't drop it down. See the video [here](https://github.com/user-attachments/assets/70c0f8e4-73fd-48b8-8977-7fbcf894b3de)

### Expected behavior

I would expect the drop down to appear and list the ID/label for each item in the `values` property.

### Screenshots

Image

### Logs

None

### Your environment

- Operating system and version (e.g., Ubuntu 20.04.2 LTS): I tested on both: macOS Tahoe 26.2 and Pop!_OS 24.04/Kernel 6.17.9-76061709 with the same result.
- Espanso version (e.g., 2.1.4-beta): 2.3.0
 

### Self-service

wellorbetter:
### Description

Add label/id object support for choice fields in form mode. When a choice field uses `{label, id}` values, the dropdown displays the `label` text to the user but outputs the `id` value on selection.

Closes #2627

### Reasoning

The form path's choice field only accepted `Vec<String>`, so `{label, id}` objects were silently dropped during parsing, resulting in an empty dropdown. The fix threads a parallel `ids` array through the entire data pipeline:

- **YAML parsing** (`extract_values`): now handles `Value::Object({id, label})` alongside plain strings
- **Data model** (`ChoiceItem`, `ChoiceValue`): new types that carry both label and id
- **JSON serialization**: emits `{"id","label"}` objects when id ≠ label, plain strings otherwise (backward compatible)
- **Rust FFI** (`ChoiceMetadata`): added `ids: Vec<String>` field, converted to C pointer array
- **C/C++ UI** (`ChoiceFieldWrapper`, `ListFieldWrapper`): `getValue()` returns `ids[selectedIndex]` when ids is present, falls back to `GetStringSelection()` otherwise

Plain string configurations are unaffected — `From<String>` ensures id == label.

### Testing

1. Configure a form with a label/id choice field → trigger → verify dropdown shows labels
2. Select an option → verify the output text contains the id, not the label
3. Configure a form with plain string choice → trigger → verify behavior unchanged
4. Configure a form with both text field and label/id choice → verify both work
5. Configure a form with a label/id list field → verify multi-select returns ids joined by separator
6. Configure a choice with missing id or label in one entry → verify it is skipped with a warning log

### Type of change

:x: **Bug fix**
:white_check_mark: **New feature** (A non-breaking change that adds functionality)
:x: **Breaking change**
:x: **Refactor**
:x: **Performance**
:x: **Style**
:x: **Docs**
:x: **Chore**

### DIFF
--- a/espanso-modulo/src/form/config.rs
+++ b/espanso-modulo/src/form/config.rs
@@ -86,15 +86,79 @@ pub struct TextFieldConfig {
     pub multiline: bool,
 }
 
+/// A choice/list value. `Plain` means label and id are the same string;
+/// `LabelId` separates the display text (`label`) from the output value (`id`).
+#[derive(Debug, Serialize, Clone)]
+pub enum ChoiceValue {
+    Plain(String),
+    LabelId { label: String, id: String },
+}
+
+impl ChoiceValue {
+    pub fn label(&self) -> &str {
+        match self {
+            ChoiceValue::Plain(s) => s,
+            ChoiceValue::LabelId { label, .. } => label,
+        }
+    }
+
+    pub fn id(&self) -> &str {
+        match self {
+            ChoiceValue::Plain(s) => s,
+            ChoiceValue::LabelId { id, .. } => id,
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for ChoiceValue {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        use serde::de;
+
+        struct ChoiceValueVisitor;
+
+        impl<'de> de::Visitor<'de> for ChoiceValueVisitor {
+            type Value = ChoiceValue;
+
+            fn expecting(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                f.write_str("a string or an object with 'id' and 'label'")
+            }
+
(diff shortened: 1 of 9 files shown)


wellorbetter (commented):
This adds label/id support for choice/list fields in forms (fixes #2627). The key change is threading a parallel ids array through the YAML→Rust→FFI→C++ pipeline. Plain string configs are unaffected.

PR #2680, opened 2026-05-03
Add description to desktop file

xuars:
Adds a description to the desktop file:

image
See [Recognized desktop entry keys](https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html)

### DIFF
--- a/espanso/src/res/linux/espanso.desktop
+++ b/espanso/src/res/linux/espanso.desktop
@@ -1,5 +1,6 @@
 [Desktop Entry]
 Name=Espanso
+Comment=Cross-platform Text Expander written in Rust
 Exec=espanso launcher
 Icon=icon
 Type=Application


espanso/src/res/linux/espanso.desktop
```diff
@@ -1,6 +1,7 @@
 [Desktop Entry]
 Name=Espanso
+Comment=Cross-platform Text Expander written in Rust
 Exec=espanso launcher
 Icon=icon
 Type=Application
-Categories=Utility;
```
xuars, line 6: Seems like the file used a Windows end of file, so my editor changed it to the Linux one. I see no problem with this change because the file is only used in Linux environments.

Archigos:
I'll fix the remaining CI related issues later, but since your file won't effect those OSs, I'll go ahead and merge it shortly. Thank you for the update.

PR #2699, opened 2026-05-31
fix: resolve #2361 — [Bug] Unable to install in Fedora Wayland

### LINKED ISSUE #2361: [Bug] Unable to install in Fedora Wayland
### Description

They do not work as expected. There is no `build-binary` task. Installing espanso is easier with Terra, as shown [here](https://github.com/espanso/espanso/issues/1948). 

This is a bug because it wasted thirty minutes of my time, and the docs do not work as expected. 

### Steps to reproduce

1. Try to install espanso on Wayland (in my case, Fedora)
2. Fail at the `build-binary` part.

### Expected behavior

The installation docs should be useful. 

### Screenshots

_No response_

### Logs

_No response_

### Your environment

- Public source code:
- Operating system and version (e.g., Ubuntu 20.04.2 LTS):
- Espanso version (e.g., 2.1.4-beta):

### Self-service

chinhkrb113:
## Summary

The reporter is on Fedora Wayland. espanso ships separate X11 and Wayland builds, and the chosen build must match the active session. If the Linux chapter doesn't make this distinction explicit, users on Wayland may build/install the X11 variant and have it silently fail. This complements the build-task fix above

Fixes #2361

## Changes

- `docs/src/ch03-03-linux.md`

## Why

`docs/src/ch03-03-linux.md`: The reporter is on Fedora Wayland. espanso ships separate X11 and Wayland builds, and the chosen build must match the active session. If the Linux chapter doesn't make this distinction explicit, users on Wayland may build/install the X11 variant and have it silently fail. This complements the build-task fix above.

### DIFF
--- a/docs/src/ch03-03-linux.md
+++ b/docs/src/ch03-03-linux.md
@@ -1,7 +1,19 @@
 # linux
 
 Espanso on Linux comes in two different flavors: one for X11 and one for Wayland.
-If you don't know which one to choose, follow [these steps to determine which one you are running](https://unix.stackexchange.com/a/325972).
+The variant you build or install **must match your active session**, otherwise Espanso may install but silently fail to work.
+
+#### Determining your session type
+
+Run the following command to check whether you are running X11 or Wayland:
+
+```console
+echo $XDG_SESSION_TYPE
+```
+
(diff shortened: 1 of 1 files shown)


AucaCoyan:
thank you!

PR #2700, opened 2026-05-31
fix: resolve #2695 — Clipboard repalces expansion

### LINKED ISSUE #2695: Clipboard repalces expansion
### Description

Sometimes expansion is replaced by content of the clipboard.

### Steps to reproduce

Not sure how to reproduce.

I see this only on Polish Wikipedia. Can't reproduce it on EN Wiki.

I have some JavaScript running on the page.

When I'm editing an article and try to use expansion, it inserts the stuff from the clipboard. I need to use the browser address bar to get the expansion.

### Expected behavior

I expect to always have my expansion as the output of my expansion.

### Screenshots

_No response_

### Logs

```
08:15:11 [worker(362172)] [WARN] Received malformed char: data provided contains an interior nul byte at byte pos 0
08:15:11 [worker(362172)] [WARN] Received malformed char: data provided contains an interior nul byte at byte pos 0
```

Created this at 15:28 CSET. Those were the last logs.

### Your environment

- Public source code:
- Environment name and version (e.g., Rust 1.59.0): rustc 1.95.0 (59807616e 2026-04-14) (Fedora 1.95.0-1.fc43)
- Operating system and version (e.g., Ubuntu 20.04.2 LTS): Fedora 43
- Espanso version (e.g., 2.1.4-beta): 2.3.0 (from Terra)

### Additional notes

Let me know if you want me to check something

### Self-service

- [x] I'd be willing to fix this bug myself. (When guided and if you're ok, if I will use an AI agent).

chinhkrb113:
## Summary

While the reporter is on Fedora, the same class of bug exists on Windows: interior nul bytes in the expansion would prematurely terminate the wide-string written to the Windows clipboard, causing truncated or stale content. Fix here for cross-platform consistency so the malformed-input handling is uniform

Fixes #2695

## Changes

- `espanso-clipboard/src/win32/mod.rs`

## Why

`espanso-clipboard/src/win32/mod.rs`: While the reporter is on Fedora, the same class of bug exists on Windows: interior nul bytes in the expansion would prematurely terminate the wide-string written to the Windows clipboard, causing truncated or stale content. Fix here for cross-platform consistency so the malformed-input handling is uniform.

### DIFF
--- a/espanso-clipboard/src/win32/mod.rs
+++ b/espanso-clipboard/src/win32/mod.rs
@@ -55,7 +55,7 @@ impl Clipboard for Win32Clipboard {
     }
 
     fn set_text(&self, text: &str, _: &ClipboardOperationOptions) -> Result<()> {
-        let string = U16CString::from_str(text)?;
+        let string = U16CString::from_str(sanitize_text(text))?;
         let native_result = unsafe { ffi::clipboard_set_text(string.as_ptr()) };
         if native_result > 0 {
             Ok(())
@@ -152,6 +152,10 @@ fn generate_html_descriptor(html: &str) -> String {
     render
 }
 
+fn sanitize_text(text: &str) -> String {
+    text.replace('\0', "")
+}
+
 #[derive(Error, Debug)]
 pub enum Win32ClipboardError {
     #[error("clipboard set operation failed")]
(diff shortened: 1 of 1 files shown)


AucaCoyan:
Now this is interesting!
I have to reproduce the bug in windows to make this PR live. I would be happier if there is an actual real scenario where a null byte can reach `set_text` instead of adding guards just in a imaginary possible scenario

PR #2701, opened 2026-05-31
fix: resolve #2673 — Max form width not applying to long choice text

### LINKED ISSUE #2673: Max form width not applying to long choice text
### Description

I've tried to apply the `max_form_width` limit to my `default.yml` configuration file but it does not constrain long text in the choice of list fields of forms. Basic text in the form wraps in the desired text width but the overall window is wider as a result of the choice text.

### Steps to reproduce

1. Set a small form size in default config
2. Create a form trigger with long input fields
3. Trigger the match
4. The espanso form appears wider than expected

### Expected behavior

Espanso should constraint the text shown in the form fields with text wrapping or an ellipse

### Screenshots

  
Image
You can see the basic paragraph text wrap after "... you" but the choice form choice text causes the window to be wider.

### Logs

_No response_

### Your environment

- Public source code:
- Operating system and version (e.g., Ubuntu 20.04.2 LTS): MacOS Tahoe 26.3.1 (a)
- Espanso version (e.g., 2.1.4-beta): 2.3.0, via homebrew

### Self-service

chinhkrb113:
## Summary

Confirm that the Rust side that builds the form metadata before crossing the FFI boundary actually reads `max_form_width` from config and passes it to the native struct. If the C++ choice control never receives a non-zero width because the value isn't plumbed through, the fix in form.cpp will have no effect. The text-wrapping behavior working today suggests it is plumbed, but this should be checked since the bug report indicates the value is partially honored

Fixes #2673

## Changes

- `espanso-modulo/src/sys/form/mod.rs`

## Why

`espanso-modulo/src/sys/form/mod.rs`: Confirm that the Rust side that builds the form metadata before crossing the FFI boundary actually reads `max_form_width` from config and passes it to the native struct. If the C++ choice control never receives a non-zero width because the value isn't plumbed through, the fix in form.cpp will have no effect. The text-wrapping behavior working today suggests it is plumbed, but this should be checked since the bug report indicates the value is partially honored.

### DIFF
--- a/espanso-modulo/src/sys/form/mod.rs
+++ b/espanso-modulo/src/sys/form/mod.rs
@@ -166,6 +166,35 @@ mod interop {
         }
     }
 
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+
+        fn base_form(max_form_width: i32, max_form_height: i32) -> types::Form {
+            types::Form {
+                title: "Test".to_string(),
+                icon: None,
+                fields: Vec::new(),
+                max_form_width,
+                max_form_height,
+            }
+        }
+
+        #[test]
+        fn max_form_dimensions_are_forwarded_to_interop() {
+            let owned: OwnedForm = base_form(640, 480).into();
+            assert_eq!(owned.interop.maxWindowWidth, 640);
+            assert_eq!(owned.interop.maxWindowHeight, 480);
+        }
+
+        #[test]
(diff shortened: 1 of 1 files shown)


AucaCoyan (changes requested):
Mmmm, I get the intent of the PR, but the code just add 2 unit tests, that both passes.
Did you maybe forgot to add something?

PR #2706, opened 2026-06-04
ci: distribute macOS build as DMG instead of ZIP

### LINKED ISSUE #2624: Mac distribution as DMG, not DMG wrapped in ZIP
### Description

There's really no good reason to zip the release dmg file. Can we simply have direct access to the DMG, or if you want to provide only a zip, then wrap the .app file directly in the zip without the dmg?

### Motivation

Just an extra step it install the app that is completely unnecessary. Make it simple.

### Alternatives

_No response_

- [ ] I'd be willing to contribute this feature to Espanso myself.

LeonidasZhak:
## Summary

Changes the macOS release artifact from a ZIP-wrapped DMG to a plain DMG file.

## Problem

Currently the macOS build creates a ZIP archive of the .app bundle, but macOS users expect a DMG for app distribution. The ZIP wrapping is unnecessary and non-standard.

## Solution

- Replace `ditto -c -k` (ZIP) with `hdiutil create` (DMG) in the release workflow
- Enable the upload step that was previously commented out

## Changes

`.github/workflows/create-release-draft.yml`:
- Create DMG using `hdiutil create -format UDZO` (compressed DMG)
- Upload DMG directly to GitHub Releases

Fixes #2624

### DIFF
--- a/.github/workflows/create-release-draft.yml
+++ b/.github/workflows/create-release-draft.yml
@@ -264,15 +264,13 @@ jobs:
             --no-default-features \
             --features native-tls,modulo
           bash ./scripts/create_bundle.sh
-      - name: Create ZIP archive
+      - name: Create DMG archive
         run: |
-          ditto -c -k --sequesterRsrc --keepParent target/mac/Espanso.app Espanso-Mac-Universal.zip
+          hdiutil create -volname "Espanso" \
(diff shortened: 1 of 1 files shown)


AucaCoyan:
Thank you! Very much appreciated!

Issue #2576, opened 2025-11-15
macOS text espansion fails due to accessibility issues(?)


mzfr:
### Description

Since I was asked to open a new issue(https://github.com/espanso/espanso/issues/2573#issuecomment-3527606737) so here is all the information about it

As I said in my old comment(https://github.com/espanso/espanso/issues/2573#issuecomment-3526880625) I had already tried with latest version of espanso(2.3.0) with a clean build and it didn't work.

After seeing this(https://github.com/espanso/espanso/issues/2573#issuecomment-3532583142) comment I tried again here is what I did

- shutdown espanso
- `killall espanso` to ensure no lingering processes
- `pgrep espanso` -- make sure espanso isn't automatically restarting itself (e.g. from the launchd agent), this should have no output
- remove Espanso from the `Privacy & Security -> Accessibility settings`
- removed Espanso from the `Login Items`. T
    - he comment(https://github.com/espanso/espanso/issues/2573#issuecomment-3532583142) mentions they had `"Auca Coyan Maillot"` in their login items but for me it was just `espanso`.

- I removed the following files/folder manually

```
rm ~/Library/LaunchAgents/com.federicoterzi.espanso.plist
rm ~/Library/Preferences/com.federicoterzi.espanso.plist
rm ~/Library/Preferences/espanso.plist
rm -rf ~/Library/Caches/espanso/
```

and just to make sure I wasn't missing something I ran the commands below as well:

```
launchctl remove com.federicoterzi.espanso
tccutil reset Accessibility com.federicoterzi.espanso
```

- install Espanso 
    - I tried with both, brew and direct download from release, so I did the above process twice
- Tried running it via terminal: `espanso service start`

```bash
espanso service start                                                                                                                        

Unable to start espanso as a service as it's not been registered.
You can either register it first with `espanso service register` or
you can run it in unmanaged mode with `espanso service start --unmanaged`

NOTE: unmanaged mode means espanso does not rely on the system service manager
      to run, but as a result, you are in charge of starting/stopping espanso
      when needed.
unable to start service: not registered as a service
```

So then I ran `open -a "espanso"`, this opened the whole provide accessibility permissions popup. After all approvals, espanso still __doesn't__ work for me.

The only way I have been able to make it work is if I leave `espanso daemon` running in a tmux/terminal window                                                                                                                

### Logs

- In `~/Library/Caches/espanso/espanso.log`

```
[...]

Issue #2530, opened 2025-10-11
Epanso 2.2.7 / 2.3.0 not working on macOS Tahoe 26.0.1


barabasz:
### Description

With 2.2.7 there is no Espanso icon in the status bar.
I can launch the search bar, though.

With 2.3.0 there is "Enable Accessibility" infinitive loop.

I have the same issue on all my macs.

I can see two kinds of errors in the log:

```
23:41:56 [worker(2258)] [INFO] secure input has been acquired, preventing espanso from working correctly. Our guess is that this is caused by 'loginwindow', but there are cases in which the detection is unreliable. Full path: /System/Library/CoreServices/loginwindow.app/Contents/MacOS/loginwindow
```

or

```
[service(6676)] [INFO] kdotool missing or not available for the current wayland DE.
```

I believe this is related to https://github.com/espanso/espanso/issues/2509 which was closed 5 days ago  - but the issue wasn't fix. 

### Steps to reproduce

1. Downloading latest binary from official website or installing via `brew install espanso`
2. Trying to run Espanso
3. Expanding not working / no Espanso icon in status bar

### Logs

```
22:49:15 [worker(2258)] [INFO] secure input has been disabled
23:41:56 [worker(2258)] [INFO] secure input has been acquired, preventing espanso from working correctly. Our guess is that this is caused by 'loginwindow', but there are cases in which the detection is unreliable. Full path: /System/Library/CoreServices/loginwindow.app/Contents/MacOS/loginwindow
23:41:57 [worker(2258)] [INFO] secure input has been disabled
23:45:49 [worker(2258)] [INFO] engine eventloop has terminated, propagating exit event...
23:45:49 [worker(2258)] [INFO] waiting for engine exit mode...
23:45:49 [worker(2258)] [INFO] exiting worker (to be restarted)
23:45:49 [daemon(1965)] [INFO] worker requested a restart, spawning a new one...
23:45:49 [daemon(1965)] [INFO] spawning the worker process...
23:45:50 [worker(4420)] [INFO] reading configs from: "/Users/barabasz/.config/espanso"
23:45:50 [worker(4420)] [INFO] reading packages from: "/Users/barabasz/.config/espanso/match/packages"
23:45:50 [worker(4420)] [INFO] using runtime dir: "/Users/barabasz/Library/Caches/espanso"
23:45:50 [worker(4420)] [INFO] system info: Darwin v26.0.1 - kernel: 25.0.0
[...]

Issue #2606, opened 2025-12-20
Espanso 2.3.0 running in issues on Windows 11 + hyphen issues


valnerwasilewski:
### Description

It's been a while that I've been running into issues with Espanso since I switched from Windows 10 to Windows 11 (since 2.2.3 and still hapenning in 2.3.0)

```
PS C:\Users\wslwsk> 14:53:04 [worker(122800)] [ERROR] thread 'engine thread' panicked at 'capacity overflow': /rustc/1159e78c4747b02ef996e55082b704c09b970588\library\core\src\iter\traits\iterator.rs:2027
14:53:04 [worker(122800)] [ERROR] thread 'detect thread' panicked at 'unable to send to the source channel: "SendError(..)"': espanso\src\cli\worker\engine\funnel\mod.rs:104
14:53:04 [worker(122800)] [ERROR] thread 'detect thread' panicked at 'panic in a function that cannot unwind': library\core\src\panicking.rs:225
thread caused non-unwinding panic. aborting.
14:53:21 [daemon(129960)] [ERROR] received unexpected exit code from worker -1073740791, exiting

thread 'main' panicked at espanso\src\cli\launcher\mod.rs:196:49:
failed to launch daemon: unexpected error, 'espanso daemon' returned a non-zero exit code.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I had all my shortcuts starting with hyphen (-). It worked great in Windows 10. However, in this new version I cannot type hyphen whenever Espanso is running. Everytime I try to use hyphen, it seems to quickly remove the hyphen I type or even didn't type it all. I've shared it on [Reddit community](https://www.reddit.com/r/espanso/comments/1nmc2xg/espanso_is_not_allowing_to_use_hyphen_in_triggers/) and they recommended to add ```win32_exclude_orphan_events: false``` to my config/default.yml but it didn't work.

Also, after a while, Espanso just stops working, and in CMD it shows that message. On Espanso logs, doesn't seem anything is different.

### Steps to reproduce

Hyphen issue:
1. Start Espanso
2. Type a hyphen
3. Hyphen doesn't work at all

Espanso crashes
1. Start Espanso
2. Start using
3. Espanso stops working with that error in CMD

### Expected behavior

Hyphen can be placed and used with Espanso
Espanso doesn't stop to work from nowhere.

### Screenshots

[...]

PR #2725, opened 2026-06-29
fix(daemon): respawn worker on unexpected exit instead of hanging or quitting

emanblue:
## Problem

On macOS (service mode), espanso can silently stop expanding for hours: the `daemon` process stays alive while the `worker` process is gone and is never respawned. `espanso service status` reports "not running" even though the daemon PID is alive, and a manual `espanso restart` is required to recover. Reported in #2530 (the dead-worker symptom also appears in #2576).

## Root cause

In `espanso/src/cli/daemon/mod.rs`:

1. `spawn_worker()`'s `worker-status-monitor` thread only forwarded an exit code to the daemon when the worker exited **non-zero** (or died by signal → `WORKER_ERROR_EXIT_NO_CODE`). A worker that exited `WORKER_SUCCESS` (0) unexpectedly — or a failed `child.wait()` — was dropped silently, so the daemon's `select!` loop was never notified and the worker was never respawned (the "half-dead" state).
2. In the `exit_signal` arm, any code other than `WORKER_EXIT_ALL_PROCESSES` / `WORKER_RESTART` hit `_ => { ...; return code; }`, so an *unexpected* worker death made the daemon **exit**. Combined with the macOS launchd plist having `RunAtLoad` but no `KeepAlive`, nothing relaunched espanso.

## Fix

- Forward **every** worker exit to the daemon (including `WORKER_SUCCESS` and the `wait()` error case), so the daemon is always notified.
- On an **unexpected** worker exit, respawn the worker with bounded, backing-off retries (`MAX_WORKER_AUTO_RESTARTS`), giving up only if it dies repeatedly in quick succession (crash-loop protection; the counter resets after a healthy run of `WORKER_HEALTHY_RUNTIME`).
- Track daemon-initiated restarts (config / keyboard-layout change) via an `expected_worker_exits` counter so the deliberately-terminated old worker is not mistaken for an unexpected death and double-spawned. `restart_worker` now returns whether it actually spawned a replacement.

## Testing

Built with `--no-default-features --features vendored-tls`; clippy clean on the changed file. Manual test on macOS with isolated config/runtime dirs:

- **Kill the worker** → daemon logs `worker exited unexpectedly with code 90` then `respawning worker (attempt 1/5) after 500ms`; a new worker starts. (Before this change: the daemon hung with no worker.)
- **Repeated rapid kills** → backoff `500ms → 1s → 1.5s → 2s → 2.5s`, then `worker died 5 times in quick succession; giving up to avoid a crash-loop` and the daemon exits cleanly (no infinite spin).
- **Normal config-change restart** still works without double-spawning (expected-exit accounting).

## Note / open question

This makes the daemon resilient to worker death regardless of cause. On macOS the trigger appears to be the worker being torn down across sleep / secure-input transitions (see #2530). Separately, the macOS launchd plist lacks `KeepAlive`; I deliberately left that out of this PR because the launchd-managed process is the short-lived `launcher`, so adding `KeepAlive` naively would cause a respawn loop — happy to address it properly in a follow-up if maintainers prefer.

### DIFF
--- a/espanso/src/cli/daemon/mod.rs
+++ b/espanso/src/cli/daemon/mod.rs
@@ -32,7 +32,7 @@ use crate::{
     common_flags::*,
     exit_code::{
         DAEMON_ALREADY_RUNNING, DAEMON_FATAL_CONFIG_ERROR, DAEMON_GENERAL_ERROR, DAEMON_SUCCESS,
-        WORKER_ERROR_EXIT_NO_CODE, WORKER_EXIT_ALL_PROCESSES, WORKER_RESTART, WORKER_SUCCESS,
+        WORKER_ERROR_EXIT_NO_CODE, WORKER_EXIT_ALL_PROCESSES, WORKER_RESTART,
     },
     ipc::{create_ipc_client_to_worker, IPCEvent},
     lock::{acquire_daemon_lock, acquire_worker_lock},
@@ -45,6 +45,15 @@ mod ipc;
 mod keyboard_layout_watcher;
 mod troubleshoot;
 mod watcher;
+
+/// Maximum number of times the daemon will automatically respawn a worker that
+/// keeps dying in quick succession before giving up, to avoid a tight
+/// crash-loop (e.g. when the worker can't start due to missing permissions).
+const MAX_WORKER_AUTO_RESTARTS: u32 = 5;
+
+/// If a worker stays alive at least this long before dying, it is considered to
+/// have run successfully and the auto-restart counter is reset.
+const WORKER_HEALTHY_RUNTIME: std::time::Duration = std::time::Duration::from_secs(60);
 
 pub fn new() -> CliModule {
     #[allow(clippy::needless_update)]
@@ -129,6 +138,16 @@ fn daemon_main(args: CliModuleArgs) -> i32 {
     ipc::initialize_and_spawn(&paths.runtime, exit_notify.clone())
         .expect("unable to initialize ipc server for daemon");
 
+    // Bookkeeping for automatic worker respawn (see issue #2530).
+    // `last_worker_spawn` tracks how long the current worker has been alive, so
+    // a healthy worker resets the crash-loop counter. `worker_auto_restart_count`
+    // counts consecutive rapid respawns. `expected_worker_exits` counts worker
+    // exits the daemon initiated itself (during a restart) and must therefore
+    // NOT be treated as unexpected deaths.
+    let mut last_worker_spawn = Instant::now();
+    let mut worker_auto_restart_count: u32 = 0;
+    let mut expected_worker_exits: u32 = 0;
+
     loop {
         select! {
           recv(watcher_signal) -> _ => {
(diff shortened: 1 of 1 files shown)


EvanAhern:
Strong +1 on this fix, with field data from macOS confirming the exact failure path this PR addresses.

**Environment:** espanso 2.3.0 (Homebrew cask), macOS 26.5 (25F71), ARM64 (MacBookPro17,1).

**What we observed:** five identical crash cascades within three days (accelerating; first occurrence 2026-06-30), each one:

```
[worker] thread 'engine thread' panicked at 'capacity overflow': .../core/src/iter/traits/iterator.rs:2027
[worker] thread 'detect thread' panicked at 'unable to send to the source channel: "SendError(..)"': espanso/src/cli/worker/engine/funnel/mod.rs:104
[worker] unable to send UIEvent to engine: sending on a disconnected channel
[worker] thread 'main' panicked at 'broken UI->Engine channel': espanso/src/cli/worker/mod.rs:150
[worker] thread 'main' panicked at 'panic in a function that cannot unwind'
thread caused non-unwinding panic. aborting.
[daemon] received unexpected exit code from worker 90, exiting
```

The engine-thread `capacity overflow` at `iterator.rs:2027` appears to be the same panic reported on Windows in #2606 (there on both 2.2.3 and 2.3.0), so the trigger looks cross-platform; on this machine it now fires every few hours while idle.

**Why this PR matters:** after each abort, the daemon exits per the code path this PR changes, and since the launchd plist that `espanso service register` writes has `RunAtLoad` but no `KeepAlive`, nothing ever relaunches espanso — every crash became a silent, indefinite, total outage of all expansions until a human noticed. With the bounded respawn in this PR, each of our five outages would have been a sub-second blip instead.

One adjacent observation while debugging (happy to file separately with full details if useful): with a lingering `espanso modulo welcome` child open, the worker's engine loop appears parked — expansions stop, IPC requests are accepted but queue undrained, and a daemon-initiated worker restart times out ("could not restart worker, as the exit process has timed out"), after which the daemon's restart machinery no longer reacts to config changes. Killing the modulo child immediately unblocks the queued events and the pending exit. That wedge is a different path than the dead-worker respawn fixed here, but it interacts with the same "daemon gives up" outcome.

Full `.ips` crash reports (two from 2026-07-06 alone) available on request. Thanks for digging into the daemon lifecycle — this fix addresses a real reliability gap.

AucaCoyan (approved):
nice job!

marcdhansen:
Tested this PR locally on Apple Silicon with macOS 26.6.1 (Darwin 25.6.0). The documented macOS Cargo build completed successfully. I launched the PR binary with isolated configuration and runtime directories, terminated only its worker, and reproduced worker exit code 90. The daemon stayed alive, logged the bounded respawn attempt, spawned a replacement worker, and the replacement worker initialized successfully.

alvanx:
Another macOS 26 report, adding it since this PR looks like the fix for the half-dead state.

**Environment:** espanso 2.4.1, macOS 26.5.2 (25F84), Apple Silicon (MacBookPro18,1).

**Two separate failures:**

1. **Worker aborts (recovers on its own).** Four crash reports between Sep 16 and 17, all with the same stack: `panic_cannot_unwind` in `espanso_ui::mac::MacEventLoop`'s `UIEventLoop::run` callback, called from `__NSFireTimer`. The worker had been running anywhere from 1 minute to 14 hours. Espanso came back by itself each time.

2. **Half-dead state (does not recover).** Twice (Sep 18 and Sep 22) the worker disappeared with **no crash report**, while the daemon stayed alive. `espanso status` reports "not running", and `espanso start` shows the "already running" window. Expansion stays dead until you intervene. On Sep 18, the last worker log lines are secure input being taken and released by `loginwindow` (a screen lock), seconds before the worker vanished. I don't have the log for Sep 22 because the restart overwrote it.

The only way to recover is to kill every espanso process, delete the `.sock`/`.lock` files in the runtime dir, then run `espanso start`.

A worker exit with no crash report plus a daemon that never notices fits the silently dropped `WORKER_SUCCESS` / failed `wait()` path this PR fixes. I haven't confirmed the exit code, though. It would be great to see this merged.

PR #2731, opened 2026-07-13
Add WaylandAppInfoProvider for wlroots-based compositors (sway, Hyprland, labwc, wayfire, etc.)

### LINKED ISSUE #2730: Add WaylandAppInfoProvider for wlroots-based compositors (labwc, sway, etc.)
### Description

Espanso currently supports Wayland app detection only for Niri (niri msg focused-window) and KDE (kdotool). On wlroots-based compositors like labwc, the log shows:

  no appropriate WaylandAppInfoProvider found for current DE/WM

  This makes filter_class, filter_title, and filter_exec non-functional, preventing per-app configs (e.g. different paste_shortcut for terminals vs browsers).

  Proposed solution:

  wlroots compositors support the wlr-foreign-toplevel-management protocol, which exposes toplevel window state including the activated flag. The wlrctl CLI tool already provides this:

  # List all windows (format: "app_id: title")
  wlrctl toplevel list

  # Check if a specific app is focused
  wlrctl toplevel find state:active app_id:Alacritty

  A WaylandWlrootsAppInfoProvider could use wlrctl (similar to how WaylandKDEAppInfoProvider uses kdotool) to query the focused window's app_id, title, and PID.

  Detection of wlroots compositors could check XDG_SESSION_DESKTOP or XDG_CURRENT_DESKTOP for known wlroots-based DEs (labwc, sway, Hyprland, Budgie 10, wayfire), or fall back to checking if wlrctl is available and functional.

  Affected compositors: labwc, sway, Hyprland, wayfire, Budgie 10 (uses labwc), and any other wlroots-based compositor.

  Environment: Fedora 44, Budgie 10 (labwc), espanso 2.2.7, wlrctl 0.2.2.

  Current workaround: A fake ~/bin/niri shell script that translates niri msg focused-window into wlrctl calls, combined with XDG_SESSION_DESKTOP=niri in the systemd service.

### Motivation

_No response_

### Alternatives

_No response_

- [x] I'd be willing to contribute this feature to Espanso myself.

pafernanr:
fixes: #2730

### DIFF
--- a/espanso-info/src/lib.rs
+++ b/espanso-info/src/lib.rs
@@ -108,7 +108,6 @@ pub fn get_provider() -> Result<Box<dyn AppInfoProvider>> {
                 }
                 info!("wlrctl missing or not available for the current wayland DE.");
             }
-            "" => continue,
             _ => {}
         }
     }
--- a/espanso-info/src/lib.rs
+++ b/espanso-info/src/lib.rs
@@ -69,16 +69,22 @@ pub fn get_provider() -> Result<Box<dyn AppInfoProvider>> {
 #[allow(clippy::match_str_case_mismatch)]
 pub fn get_provider() -> Result<Box<dyn AppInfoProvider>> {
     use std::env;
+    use std::process::Command;
 
-    if let Ok(value) = env::var("XDG_SESSION_DESKTOP") {
-        match value.to_lowercase().as_str() {
+    let session_desktop = env::var("XDG_SESSION_DESKTOP")
+        .unwrap_or_default()
+        .to_lowercase();
+    let current_desktop = env::var("XDG_CURRENT_DESKTOP")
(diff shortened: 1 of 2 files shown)


pafernanr:
if wlrctl exposed the PID, we could resolve exec via readlink /proc/<pid>/exe, which would enable filter_exec (exactly what the Niri and KDE providers do). Without it, only filter_class and filter_title work.

That said, the underlying wlr-foreign-toplevel-management Wayland protocol doesn't include PID information either, so wlrctl can't really expose what the protocol doesn't provide. It would require compositor-side extensions or a different protocol to get PIDs.

For most users filter_class (app_id) is sufficient — it's the primary way people write per-app configs. So this is a reasonable first implementation, and PID support could be added later if a mechanism becomes available.

PR #2735, opened 2026-07-18
make clippy check only our deps

AucaCoyan:
In order to make the CI faster, I believe if we can make clippy check only our deps (`espanso-*`) it can significantly reduce the time


AucaCoyan:
it's not taking less time, so... no

Issue #2509, opened 2025-10-06
"kdotool missing or not available" on macOS


barabasz:
### Description

I'm unable to start espanso on macOS Tahoe after I unchecked "App Background Activity" (I didn't know that "Auca Coyan Maillot" equals "Espanso"...). Now, even if I re-enable it again, espanso is not working. Reinstalling does not help.

![Image](https://github.com/user-attachments/assets/b996056f-880e-40b1-9430-9105119dee7e)

### Steps to reproduce

1. download [latest ZIP](https://github.com/espanso/espanso/releases/latest/download/Espanso-Mac-Universal.zip) or install via `brew install espanso`
2. try to start espanso

### Expected behavior

Espanso running in the background. 

### Logs

```
00:35:41 [service(2253)] [INFO] reading configs from: "/Users/barabasz/.config/espanso"
00:35:41 [service(2253)] [INFO] reading packages from: "/Users/barabasz/.config/espanso/match/packages"
00:35:41 [service(2253)] [INFO] using runtime dir: "/Users/barabasz/Library/Caches/espanso"
00:35:41 [service(2253)] [INFO] system info: Darwin v26.0.1 - kernel: 25.0.0
00:35:41 [service(2253)] [INFO] kdotool missing or not available for the current wayland DE.
00:35:41 [service(2253)] [INFO] reloading espanso launchctl entry
00:35:41 [service(2253)] [INFO] service registered correctly!
00:36:08 [service(2365)] [INFO] reading configs from: "/Users/barabasz/.config/espanso"
00:36:08 [service(2365)] [INFO] reading packages from: "/Users/barabasz/.config/espanso/match/packages"
00:36:08 [service(2365)] [INFO] using runtime dir: "/Users/barabasz/Library/Caches/espanso"
00:36:08 [service(2365)] [INFO] system info: Darwin v26.0.1 - kernel: 25.0.0
00:36:08 [service(2365)] [INFO] kdotool missing or not available for the current wayland DE.
00:36:13 [service(2365)] [ERROR] unable to start service: timed out
00:36:13 [service(2365)] [ERROR] Hint: sometimes this happens because another Espanso process is left running for some reason.
00:36:13 [service(2365)] [ERROR]       Please try running 'espanso restart' or manually killing all Espanso processes, then try again.
```

### Your environment

- Operating system and version: Tahoe 26.0.1 
- Espanso version: 2.2.7

### Self-service

- [ ] I'd be willing to fix this bug myself.

[...]

Issue #2162, opened 2024-11-30
Espanso doesn't work with certain applications on Wayland (KDE Plasma, GTK)


mathe00:
### Description

For several months, I have been encountering difficulties getting Espanso to work properly on my system. It works in some applications (such as Electron-based apps like Brave and Discord), but not in other native KDE Plasma applications like Dolphin (file explorer) or Konsole (terminal). I suspect the issue may be related to Qt-based applications or a specific interaction with Wayland. I’ve tried several solutions suggested in the documentation and on forums without success.

I have also attempted various changes to the settings, including trying out what seemed to be a similar issue reported in **#768** without any improvement. 

I’ve checked Espanso’s configuration, and the **default.yml** file appears correct (see excerpt below):

```yaml
➜  ~ cat /home/temo/.config/espanso/config/default.yml
label: null
backend: null
enable: null
clipboard_threshold: null
auto_restart: null
preserve_clipboard: null
toggle_key: OFF
paste_shortcut: null
disable_x11_fast_inject: null
word_separators: null
backspace_limit: null
apply_patch: null
search_trigger: null
search_shortcut: 'CTRL+SPACE'
undo_backspace: true
show_notifications: null
show_icon: null
secure_input_notification: null
post_form_delay: null
post_search_delay: null
emulate_alt_codes: null
win32_exclude_orphan_events: null
win32_keyboard_layout_cache_interval: null
x11_use_xclip_backend: null
x11_use_xdotool_backend: null
pre_paste_delay: null
restore_clipboard_delay: null
paste_shortcut_event_delay: null
inject_delay: null
key_delay: null
keyboard_layout:
  layout: "fr"
  variant: "azerty"
  model: "pc105"
evdev_modifier_delay: null
includes: null
excludes: null
extra_includes: null
extra_excludes: null
use_standard_includes: null
filter_title: null
filter_class: null
filter_exec: null
filter_os: null
```

Despite these efforts, Espanso still doesn’t work as expected in Qt-based KDE Plasma applications. Any advice or further troubleshooting steps would be greatly appreciated. 

### Steps to reproduce

1. Install Espanso on a system running KDE Plasma with Wayland (Arch Linux, Plasma 6.2.0).
2. Launch applications like Dolphin or Konsole.
3. Try using the shortcuts defined in Espanso (e.g., text replacement or abbreviation).
4. Observe that the shortcuts don't work as expected in these applications (however, they work in apps like Brave or Discord).

### Expected behavior

[...]

PR #2741, opened 2026-07-19
feat(wayland): detect active window on KDE natively via KWin D-Bus sc…

Nimamoh:
On KDE Plasma Wayland, espanso relied on the external kdotool binary (an optional PATH dependency users had to install separately) to detect the active window for app-specific matching. When missing, app-specific configs silently stopped working (#2162, #2366, #2509).

kdotool itself talks to KWin's scripting D-Bus interface: it loads a small KWin script that reads workspace.activeWindow and reports back via callDBus() addressed at the caller's unique bus name. This commit implements that mechanism directly in espanso-info (zbus, pure Rust):

- espanso-info/src/wayland/kde_dbus.rs: KwinScriptingClient owns a session-bus connection and a receiver thread (created once per provider, not per query); a sequence number in the JSON payload guards against stale replies from timed-out queries.
- WaylandKDEAppInfoProvider now queries title/class/pid in a single round trip (previously 3 kdotool subprocesses + readlink per query) and no longer panics on non-UTF8 output.
- Provider selection probes NameHasOwner("org.kde.KWin") instead of spawning kdotool; the duplicate startup probe in main.rs (which also ran on X11 builds and caused log spam, #2366/#2509) is removed.
- zbus is MIT-licensed, already in Cargo.lock (via notify-rust), and scoped to Linux + the wayland feature, so X11/macOS/Windows builds are unaffected.

Like kdotool, this requires KWin's scripting interface (Plasma 5/6); tested against KWin 6.7.3 (kwin_wayland --virtual) with XWayland and native-Wayland clients, including app-specific filter_class matching, with kdotool absent from the system.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1044,7 +1044,9 @@ dependencies = [
  "cc",
  "espanso-package",
  "log",
+ "serde_json",
  "widestring",
+ "zbus",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,7 @@ tempdir = "0.3.7"
 thiserror = "1.0.23"
 walkdir = "2.3.1"
 widestring = "0.4.3"
+zbus = "5.9.0"
 windows = { version = "0.61.1", features = [
   "Win32_UI_WindowsAndMessaging",
   "Win32_System_Console",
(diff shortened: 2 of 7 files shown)


AucaCoyan:
Thank you! this looks great!

AucaCoyan (approved):

PR #2749, opened 2026-07-22
fix(daemon/macos): take over stale daemon lock after logout/login

### LINKED ISSUE #1492: macOS logout/login doesn't restart cleanly
**Describe the bug**
Logging out and then logging in to user with espanso installed fails to restart cleanly.

**To Reproduce**
1. Reboot macOS.
2. Login to user with espanso installed. ps shows:
Screenshot 2022-12-30 at 19 56 16

3. Logout
4. Login again.
5. espanso displays the following modal:
Screenshot 2022-12-30 at 19 57 25

6. pse shows:
Screenshot 2022-12-30 at 19 57 45

7. To clean things up, it's necessary to kill -9 the process and and reload espanso

**Expected behavior**
espanso should unload all process on logout and restart cleanly on login.

**Setup information**
- OS: macOS 12.6.2 (Monterey)
- Version: espanso 2.1.8

robwilkerson:
## Problem

As reported in #1492.

On macOS, espanso can appear to be running after login, but expansions don't work. I assume it would apply to fast user switch as well, but did not verify this myself. `espanso status` reports a running instance, but nothing expands until the process 
is manually killed and restarted.

A `daemon` process from the previous login session seems to survive into the new session. It holds the daemon lock, but isn't serving the current session. When the new session starts espanso, the new daemon can't acquire the lock, logs `daemon is already running!`, and exits. Because the stale daemon is still there, repeated logins leave several dead espanso processes behind as noted in the issue.

## Root cause

The daemon lock (`espanso/src/lock.rs`) is an advisory `flock` held by an open file descriptor, so the OS releases it automatically when the holding process dies. A lock that is still held implies a *live* process is holding it.

On macOS, the previous session's `daemon` is not reliably torn down on logout, so it stays alive into the next session and keeps holding the lock. In `daemon_main` (`espanso/src/cli/daemon/mod.rs`), the startup path did:

    let lock_file = acquire_daemon_lock(&paths.runtime);
    if lock_file.is_none() {
        error!("daemon is already running!");
        return DAEMON_ALREADY_RUNNING;
    }

So the new session's daemon could never acquire the lock from the stale holder: it returned `DAEMON_ALREADY_RUNNING` and exited. Nothing tried to reclaim the lock or to shut the stale daemon down so dead processes may accumulate.

## Fix

On macOS, when the daemon lock is already held, the new daemon now tries to take over from the (presumed stale) holder instead of giving up. It:

- Asks the existing daemon to exit over IPC (`IPCEvent::Exit`) and wait up to 3s for it to release the lock, then acquire it. A live, responsive daemon shuts down cleanly and hands off.
- If the holder is wedged and does not release in time, it clears the stale lock file and acquires a fresh one. The stale process keeps its now-unlinked lock file and no longer blocks startup.
- The daemon's existing `terminate_worker_if_already_running` step then shuts down the old session's worker, so no daemon/worker processes accumulate across relogins.

I tried to mirror the existing `terminate_worker_if_already_running` pattern (same 3s / 200ms poll), applied to the daemon instead of the worker.

The change is gated behind `#[cfg(target_os = "macos")]`, so Linux and Windows startup behavior is unchanged. Supporting additions:

- `lock::clear_daemon_lock` (removes the stale daemon lock file) and a small `lock_file_path` helper / name-constant refactor so the path is defined once.
- `ipc::create_ipc_client_to_daemon`, mirroring the existing `create_ipc_client_to_worker`.

## Testing

A unit test in `lock.rs` covers the take-over mechanism: an orphan holds the lock, and clearing the stale lock file lets a new lock be acquired while the orphan's now-unlinked file is still held.

Manual test on macOS with two real `espanso daemon` processes against isolated `--config_dir` / `--runtime_dir` dirs:

- **Responsive orphan (graceful):** daemon A holds the lock; starting daemon B logs `daemon lock is held, attempting to take over from a possibly stale daemon...` then proceeds. Daemon A exits and its worker self-exits (`detected unexpected daemon termination... exiting worker process`). Result: exactly one daemon + one worker, no leftovers.
- **Wedged orphan (fallback):** daemon A frozen with `SIGSTOP` so it can't answer IPC; starting daemon B waits ~3s, logs `stale daemon did not release the lock, clearing the stale lock file`, then takes over.
- In both cases the new daemon never prints `daemon is already running!`.

Not exercised: a literal macOS logout/login cycle. The two-process test drives the exact code path a relogin hits (lock held by a live daemon → take over).

On CI: out of an abundance of caution I did not enable GitHub Actions on my fork, so the workflows have not run against these commits. I reproduced the `macos` job locally on Apple Silicon:

* `cargo fmt --all -- --check`
* `cargo test --workspace`
* `cargo clippy -- --deny warnings` 

All passed. 

I did not verify other platforms. I'll leave that to CI.

## Note

As mentioned, this changes is scoped intentionally to macOS. The bug is macOS-specific (the previous session's daemon surviving a logout/login), reproduction needs a Mac, and gating keeps Linux/Windows startup untouched. I'll be happy to generalize the mechanism if maintainers would prefer a unified, cross-platform fix.

Related: #2725 also touches `daemon/mod.rs` and reworks the worker-exit handling. This change appears independent (a different bug, and a different region of the file: lock acquisition + a new take-over helper), so I don't anticipate any conflict on `dev`, but whichever lands second may need a trivial rebase.

While verifying, I noticed a pre-existing, separate quirk: a graceful daemon shutdown over IPC logs `received unexpected exit code from worker 0, exiting` at ERROR level even though it exits cleanly. It sits in the same exit-handling code #2725 is reworking, so I left it out of this PR.

Closes #1492.

### DIFF
--- a/espanso/src/cli/daemon/mod.rs
+++ b/espanso/src/cli/daemon/mod.rs
@@ -68,6 +68,14 @@ fn daemon_main(args: CliModuleArgs) -> i32 {
 
     // Make sure only one instance of the daemon is running
     let lock_file = acquire_daemon_lock(&paths.runtime);
+
+    // On macOS, a daemon from a previous login session can survive a
+    // logout/login (see GitHub issue #1492), keeping the daemon lock held while
+    // no longer serving the current session. In that case, try to take over
+    // from the stale daemon instead of giving up.
+    #[cfg(target_os = "macos")]
+    let lock_file = lock_file.or_else(|| reclaim_stale_daemon_lock(&paths.runtime));
+
     if lock_file.is_none() {
         error!("daemon is already running!");
         return DAEMON_ALREADY_RUNNING;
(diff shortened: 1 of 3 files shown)


AucaCoyan (approved):
Looks great!

AucaCoyan:
Thank you!

PR #2532, opened 2025-10-12
Replace `serde_yaml` with `serde_norway` (#1937)

gcailly:
This PR migrates from the deprecated `serde_yaml` to the actively maintained `serde_norway`.

- Updated workspace dependency: `serde_yaml = "0.8.17"` → `serde_norway = "0.9.42"`.
- Replaced all `serde_yaml` imports with `serde_norway` across the codebase (15 files).
- Added handling for `serde_norway::Value::Tagged` variant (new feature for YAML custom tags).

Fixes #1937

PR #2761, opened 2026-08-08
fix(config): parse unquoted special-char values in YAML flow collections

### LINKED ISSUE #2748: v2.4.0 doesn't parse unquoted values in `- triggers:`
### Description

I'm trying the new release, v2.4.0.

It is throwing errors at most of my `- triggers:` lines:
```
Caused by:
    0: failed to parse YAML match group
    1: did not find expected node content at line 35 column 16, while parsing a flow node
```
The only way I can get round it is by quoting the values within the square brackets, e.g. replacing `- triggers: [:->,:>-]` with `- triggers: [':->',':>-']`. It's a bit of a pain as I have a lot. Is this behaviour intentional?

DeepWiki suggests that "The error originates from YAMLMatchGroup::parse_from_file() which calls serde_norway::from_str() to parse YAML content. parse.rs:55-58".

Are others seeing the same? It wasn't a problem with the previous versions.

### Your environment

- Public source code:
- Operating system and version (e.g., Ubuntu 20.04.2 LTS): Linux Mint 21.2 x86_64 X11
- Espanso version (e.g., 2.1.4-beta): 2.4.0

### Self-service

YuriNachos:
Fixes #2748.

## Root cause
PR #2532 swapped `serde_yaml` 0.8 → `serde_norway` (libyaml-based, strict). A plain (unquoted) scalar inside a flow collection (`[...]` / `{...}`) must not begin with a flow-indicator character (`:`, `>`, `|`); `serde_yaml` tolerated these, so configs that parsed before 2.4.0 — e.g. `triggers: [:->,:>-]` or `word_separators: [:>-]` — now fail with *"did not find expected node content ... while parsing a flow node"*.

## Fix
New `parse_lenient` parses strictly first; only on failure does it retry on a leniently re-quoted copy (`lenient_flow_quote`), which single-quotes bare flow-collection scalars that begin with `:`, `>`, or `|`. The retry-on-failure gate is the central safety property: every config that parses today is returned byte-for-byte untouched, and if the retry still fails the **original** error is returned — so genuinely malformed YAML fails exactly as before. Already-quoted scalars and `#` comments are skipped verbatim; feature/reserved indicators (`*` `!` `&` `@` `%`) are deliberately left strict (pre-2.4.0 did not accept them either).

Wired into both YAML entry points: `config/parse/yaml.rs` and `matches/group/loader/yaml/parse.rs`.

## Test plan
- `cargo fmt --all -- --check` — clean
- `cargo clippy -p espanso-config --all-targets -- --deny warnings` — clean
- `cargo test -p espanso-config` — **95 passed, 0 failed**, including: the #2748 reproducer (`:->`, `:>-`, `:-)`, `:>hello`), a zero-regression corpus of real espanso match/config shapes (proven value-identical strict vs lenient), and pathological-input robustness (unterminated quotes, NUL, 100k-deep nesting — all terminate).

### DIFF
--- a/espanso-config/src/config/parse/yaml.rs
+++ b/espanso-config/src/config/parse/yaml.rs
@@ -22,7 +22,7 @@ use serde::{Deserialize, Serialize};
 use serde_norway::Mapping;
 use std::convert::TryFrom;
 
-use crate::util::is_yaml_empty;
+use crate::util::{is_yaml_empty, parse_lenient};
 
 use super::ParsedConfig;
 
@@ -188,7 +188,7 @@ impl YAMLConfig {
             )?);
         }
 
-        Ok(serde_norway::from_str(yaml)?)
+        Ok(parse_lenient(yaml)?)
     }
 }
 
@@ -270,6 +270,20 @@ impl TryFrom<YAMLConfig> for ParsedConfig {
 mod tests {
     use super::*;
     use std::{collections::BTreeMap, convert::TryInto};
+
+    // Sibling regression for #2748 on the config-parser path: the same
+    // serde_yaml -> serde_norway flow-indicator rule (PR #2532) also rejects
+    // unquoted flow values in config YAML, e.g. an indicator-led `word_separators`
+    // entry. It must parse after the fix.
+    #[test]
+    fn parse_unquoted_flow_value_in_config_issue_2748() {
+        let yaml = "word_separators: [:->, :>-]\n";
+        let config = YAMLConfig::parse_from_str(yaml).unwrap();
+        assert_eq!(
+            config.word_separators,
+            Some(vec![":->".to_string(), ":>-".to_string()])
+        );
+    }
 
     #[test]
     fn conversion_to_parsed_config_works_correctly() {
--- a/espanso-config/src/matches/group/loader/yaml/parse.rs
+++ b/espanso-config/src/matches/group/loader/yaml/parse.rs
@@ -23,7 +23,7 @@ use anyhow::Result;
 use serde::{Deserialize, Serialize};
 use serde_norway::Mapping;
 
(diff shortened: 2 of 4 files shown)


YuriNachos:
Updated to fix a defect a post-publish adversarial review found.

**Before:** `lenient_flow_quote` re-quoted inside `|` / `>` block-scalar bodies. With a config that combined a #2748 trigger and a `replace: |` whose literal body contained flow-like text, the body was silently corrupted — e.g. `if [ :>x ]; then echo hi; fi` became `if [ ':>x' ]; then echo hi; fi` (broken shell command). That is a worse failure mode than the 2.4.0 regression this PR fixes.

**Now:** the transform detects `|` / `>` block-scalar headers in block context (`depth == 0`) and copies the whole block — header + body — verbatim, so literal text in a block scalar is never mistaken for YAML flow structure. Detection is deliberately liberal (a false positive only misses a fix in that block; a false negative could corrupt data). Added a regression test combining the #2748 trigger with a `replace: |` body containing `[ :>x ]`, asserting the body is byte-identical.

`cargo test -p espanso-config` — **97 passed, 0 failed** (was 95). Code change confined to `espanso-config/src/util.rs` (+ the test in `parse.rs`); the strict-first / retry-on-failure / original-error-preserved safety properties are unchanged.

YuriNachos:
Since this one repairs the 2.4.0 flow-collection parsing regression from the serde_norway swap (#2748), I'd appreciate a sanity check on the approach: the config is still parsed strictly first, and the leniently re-quoted retry only runs when that strict parse fails. Would you be open to reviewing it as-is, or is there a different layering or additional regression cases you'd want covered first? I'm glad to prioritise whatever would help it move forward.

AucaCoyan:
Thank you so much!

PR #2763, opened 2026-08-09
Fix intermittent plain v on macOS clipboard expansion

### LINKED ISSUE #1288: When pasting a snippet sometimes "v" is pasted instead of the acrual snippet
**Describe the bug**
When pasting a snippet sometimes "v" is pasted instead of the actual snippet

**Expected behavior**
The actual snippet is pasted.

**Setup information**
- OS: MacOS
- Version:  2.1.6-beta

rsuzuki0:
## Summary

- attach explicit Quartz modifier flags to every synthesized key event in a virtual-key combination
- preserve the active modifier state while pressing and releasing the combination
- prevent an intended Command-V paste from intermittently arriving as a plain `v` on macOS

Fixes #1288.

## Root cause

`inject_vkeys_combination` posted the modifier key-down and the non-modifier key-down as separate events, but did not set `CGEventFlags` on either event. Under some timing conditions, the target application could process the V event without the Command modifier and insert a literal `v`.

The change derives Quartz flags from the virtual modifier keys, accumulates them during key-down, and removes them during reverse-order key-up. The V key-down and key-up therefore explicitly carry `kCGEventFlagMaskCommand` during a Command-V combination.

## User impact

Clipboard-mode expansions retain their fast insertion behavior while avoiding the intermittent lone-`v` failure. This removes the need to work around the issue with slower character-by-character injection.

## Validation

- `cargo test -p espanso-inject` — 3 passed
- `cargo clippy -p espanso-inject --all-targets -- -D warnings` — passed
- `cargo fmt --all -- --check` — passed
- Objective-C++ `-fsyntax-only` check with warnings treated as errors, excluding two pre-existing sign-comparison warnings — passed
- manual Apple Silicon field test on macOS 14.8.9 with Espanso 2.4.0 across MailMate, iTerm2, Safari, and native Save As dialogs: repeated long and short clipboard expansions produced no lone `v` failures after the unpatched path had failed frequently; behavior remained clean after configuration reloads

### DIFF
--- a/espanso-inject/src/mac/native.mm
+++ b/espanso-inject/src/mac/native.mm
@@ -27,6 +27,32 @@
 // so that we can later skip them in the detect module.
 CGPoint ESPANSO_POINT_MARKER = CGPointMake(-27469, 0);
 
+// Return the Quartz modifier flag associated with a macOS virtual key code.
+// A combination's non-modifier events need explicit flags: relying only on a
+// separately posted modifier key-down is racy and can turn CMD+V into plain V
+// (https://github.com/espanso/espanso/issues/1288).
+static CGEventFlags modifier_flag_for_vkey(int32_t vkey)
+{
+  switch (vkey) {
+    case 0x36: // Right Command
+    case 0x37: // Left Command
+      return kCGEventFlagMaskCommand;
+    case 0x38: // Left Shift
+    case 0x3C: // Right Shift
+      return kCGEventFlagMaskShift;
+    case 0x3A: // Left Option
+    case 0x3D: // Right Option
+      return kCGEventFlagMaskAlternate;
+    case 0x3B: // Left Control
+    case 0x3E: // Right Control
+      return kCGEventFlagMaskControl;
+    case 0x39: // Caps Lock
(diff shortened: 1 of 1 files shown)


mil84:
Great job! This is easily the most critical bug fix espanso needs. It would be incredibly valuable for *all Mac users* who had to deal with this annoying bug for years.

Is there any indication of when it will be reviewed/released @AucaCoyan?

rsuzuki0:
Since my fix, I haven't seen a single incident of "v" on my Mac in a few days. I used to see this problem many times per hour of working time. The v is gone.

AucaCoyan:
Hi! Sorry for the delay! Awesome solution. I'll merge right away if the CI passes! 🙌🏼

AucaCoyan (approved):

mil84:
@AucaCoyan I noticed this fix wasn't included in the recent 2.4.1 release. Do you have any estimate on when it might make it into a future version? Will this be included in the near-future 2.4.x release, or is it planned for a more distant version like 2.5 or beyond? Sorry to bother you, but this bug feels like a critical issue for most Mac users...Thanks 🙏

Issue #2668, opened 2026-04-09
Regex Trigger doesn't match after n characters or doesn't replace entire string (could be two issues, unsure)


leeand00:
### Description

Espanso version 2.3.0

Here is the regex match and the configuration file.   I am not seeing any errors in the logs, which I have also included below.

```yml
matches:
  - regex: ":tw5tabs\\((?P<whattag>.*)\\)"
    replace: "<<tabs \"[tag[{{whattag}}]sort[order]]\" >>"
...
```
`$CONFIG/espanso/match/tw5.yml`

```yml
# espanso configuration file
keyboard_layout: 
  layout: us
#  rules: "evdev"
  model: "pc105"

# toggle_key: RIGHT_CTRL

# yaml-language-server: $schema=https://raw.githubusercontent.com/espanso/espanso/dev/schemas/config.schema.json

# For a complete introduction, visit the official docs at: https://espanso.org/docs/

# You can use this file to define the global configuration options for espanso.
# These are the parameters that will be used by default on every application,
# but you can also override them on a per-application basis.

# To make customization easier, this file contains some of the commonly used
# parameters. Feel free to uncomment and tune them to fit your needs!

# --- Toggle key

# Customize the key used to disable and enable espanso (when double tapped)
# Available options: CTRL, SHIFT, ALT, CMD, OFF
# You can also specify the key variant, such as LEFT_CTRL, RIGHT_SHIFT, etc...
# toggle_key: ALT
# You can also disable the toggle key completely with
# toggle_key: OFF

# --- Injection Backend

# Espanso supports multiple ways of injecting text into applications. Each of
# them has its quirks, therefore you may want to change it if you are having problems.
# By default, espanso uses the "Auto" backend which should work well in most cases,
# but you may want to try the "Clipboard" or "Inject" backend in case of issues.
backend: Auto

# --- Auto-restart

# Enable/disable the config auto-reload after a file change is detected.
# auto_restart: false

# --- Clipboard threshold

# Because injecting long texts char-by-char is a slow operation, espanso automatically
# uses the clipboard if the text is longer than 'clipboard_threshold' characters.
clipboard_threshold: 1024

key_delay: 50
inject_delay: 50
max_regex_buffer_size: 1024
# For a list of all the available options, visit the official docs at: https://espanso.org/docs/
```
`$CONFIG/espanso/default.yml`

```
$ journalctl --user --unit espanso.service
...
[...]

PR #2764, opened 2026-08-11
fix(match/regex): cap buffer when a single event delivers multiple chars

YuriNachos:
## Problem

The `RegexMatcher` keeps a rolling buffer of recently typed characters (capped at `max_buffer_size`, default 30) and runs the user's regexes against it. The cap was enforced like this:

```rust
if buffer.len() > self.max_buffer_size {
    buffer.remove(0);
}
```

This trims **one** character per event. But an `Event::Key` can carry several characters at once (composed-key input, IME commits, paste-as-keys). When such an event arrives, the buffer grows by N characters in a single step while only one is dropped, so it stays well above `max_buffer_size` for many subsequent keystrokes — carrying stale content that causes missed and spurious regex matches.

This is the mechanism behind the "doesn't match after n characters" symptoms reported in #2668.

## Root cause

The buffer-length invariant (`buffer.len() <= max_buffer_size` after `process`) is only restored one character at a time, so any multi-character event violates it.

## Fix

Trim in a loop so the invariant is restored within the same event (`espanso-match/src/regex/mod.rs`):

```rust
while buffer.len() > self.max_buffer_size {
    buffer.remove(0);
}
```

Removal at index 0 always operates on a valid char boundary, so this stays UTF-8 safe.

## Test plan

Added `matcher_truncates_buffer_after_multichar_event`: feeds a single `Event::Key` carrying 10 characters into a matcher with `max_buffer_size: 5` and asserts the returned state's buffer is capped to `"fghij"` (the most recent 5). Red before the loop fix (`buffer` stayed `"abcdefghij"`, length 10), green after.

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p espanso-match --all-targets -- --deny warnings` — clean
- `cargo test -p espanso-match` — 18 passed, 0 failed

Authored on macOS arm64; the touched code is pure Rust with no platform conditionals, so the Linux X11/Wayland and Windows CI jobs exercise the identical path.

### DIFF
--- a/espanso-match/src/regex/mod.rs
+++ b/espanso-match/src/regex/mod.rs
@@ -92,7 +92,13 @@ where
         }
 
         // Keep the buffer length in check
-        if buffer.len() > self.max_buffer_size {
+        // A single event can deliver several characters at once (for example a
+        // composed-key or pasted input), so trimming just one character can
+        // leave the buffer well above `max_buffer_size` for many subsequent
+        // events, holding stale content and causing missed/spurious matches.
+        // `remove(0)` always operates on a valid char boundary (index 0), so
+        // this stays UTF-8 safe while shrinking the buffer back under the cap.
+        while buffer.len() > self.max_buffer_size {
             buffer.remove(0);
         }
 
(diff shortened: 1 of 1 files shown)


YuriNachos:
Just a gentle check-in on this one — is there anything you'd like me to adjust or expand (extra test cases, a different cap strategy) to make it easier to review? Happy to do the work.

PR #2767, opened 2026-08-14
linux: FHS compliance

### LINKED ISSUE #2400: Where to locate the final binary according to Filesystem Hierarchy Standard (FHS)?
Mte90 raised an interesting point here:

> also on debian the path is `/usr/bin/espanso`

 _Originally posted by @Mte90 in [#2214](https://github.com/espanso/espanso/issues/2214#issuecomment-3079576920)_

I have a bunch of reasons the two places we can leave the binary. First, the sources of truth I have found so far:

## Documentation
For `/usr/bin`:
- [Official Debian packaging guide about `/usr/bin`](https://www.debian.org/doc/packaging-manuals/fhs/fhs-3.0.html#usrbinMostUserCommands)
- [Baeldung Linux about `/usr/bin`](https://www.baeldung.com/linux/differences-bin-sbin-usr-local#3-usrbin) alternative source
- [Blogpost about a `/usr/bin`](https://guillermodotn.github.io/posts/Where_to_manually_install_your_programs_on_linux/#1-bin-and-usrbin)

For `/usr/local/bin` (the current location up to this date):
- [Official Debian packaging guide about `/usr/local`](https://www.debian.org/doc/packaging-manuals/fhs/fhs-3.0.html#usrlocalLocalHierarchy)
- [Baeldung Linux about `/usr/local/bin`](https://www.baeldung.com/linux/differences-bin-sbin-usr-local#5-usrlocalbin) alternative source
- [Blogpost about a `/usr/local/bin`](https://guillermodotn.github.io/posts/Where_to_manually_install_your_programs_on_linux/#3-usrlocalbin)

There is also `~/.local/bin` but that is suggested for installations that do not require root privileges to install (like those you extract and run the binary)

## Questions
The suggested way to user based binaries in `/usr/bin`, but also that is favourable when you use the default package manager (apt in this case). We do not distribute espanso via the official debian repositories, but by downloading the .deb artifact (or appimage, in which case it doesn't matter the location for the FHS). The management of the `usr/bin` folder is for the distro, not by manually installing the package (like we do)

- Debian use `/usr/bin`, but do all distros do that? Do we break other compliance by moving the binary?
- Does it make sense to move the binary to `/usr/bin`, if it's possible to unwanted changes caused by `apt`?
- Shouldn't we use `/usr/local/bin` for _manually_ installed binaries, in spite of using the apt tool to install it?

## Limitations

I don't want to add logic to install the binary on different locations depending on the distro. It's going too far for a simple subject.

xuars:
With commit f07bd04 all packages now install to /usr/bin, leaving /usr/local/bin only for manual compilation. 
Now `espanso env-path` will use the user's local binaries directory, so it can be added to path without root on manual compilation.
Fixes #2400

### DIFF
--- a/espanso/src/path/linux.rs
+++ b/espanso/src/path/linux.rs
@@ -23,15 +23,16 @@ use std::path::PathBuf;
 use thiserror::Error;
 
 pub fn is_espanso_in_path() -> bool {
-    PathBuf::from("/usr/local/bin/espanso").is_file()
+    PathBuf::from("/usr/bin/espanso").is_file() || PathBuf::from("/usr/local/bin/espanso").is_file()
 }
 
 pub fn add_espanso_to_path(_: bool) -> Result<()> {
-    let target_link_dir = PathBuf::from("/usr/local/bin");
+    let target_link_dir = dirs::executable_dir().expect("Could not get executables directory");
+
     let exec_path = get_binary_path()?;
 
     if !target_link_dir.is_dir() {
-        return Err(PathError::UsrLocalBinDirDoesNotExist.into());
+        return Err(PathError::LocalBinDirDoesNotExist.into());
     }
 
     let target_link_path = target_link_dir.join("espanso");
@@ -42,7 +43,7 @@ pub fn add_espanso_to_path(_: bool) -> Result<()> {
     }
 
     if let Err(error) = std::os::unix::fs::symlink(exec_path, target_link_path) {
-        debug!("creating the symlink from executable to /usr/local/bin");
+        debug!("creating the symlink from executable to $HOME/.local/bin");
         return Err(PathError::SymlinkError(error).into());
     }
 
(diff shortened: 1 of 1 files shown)


AucaCoyan:
> all packages now install to /usr/bin, leaving /usr/local/bin only for manual compilation

Hi! I consulted in Discord [community](https://discord.com/channels/884163483409731584/1013902954350248008/1537847391263395921) if that folder was right, and `/usr/local/bin` shouldn't be touched, because package managers of the distros normally install on the `/usr/bin` folder. Now, Distros don't break folders easily, but I don't see a sustable reason to change the install folder of espanso.

If commit https://github.com/espanso/espanso/commit/f07bd0472963f8b357b7d27e7f6f21c675ccc090 does install there, then I did a poor job reviewing the PR, should have catched this earlier.

Give me a bit more time to check with another colleague and maybe we can workout an easier solution 🙏🏼

xuars:
I am very sorry! On https://github.com/espanso/espanso/commit/f07bd0472963f8b357b7d27e7f6f21c675ccc090#diff-b9b872cf52f4e68169da8d1ec848cdabbd530d2043503963e75d366f2503b422R111 I touched the install location in rpm. I should have mentioned it in commit message. 

After that PR got merged, I found this issue and by reading the comments thought the change in my previous commit was part of it and I could close it with this MR.

@AucaCoyan Please feel free to revert the commit.

PR #2768, opened 2026-08-14
fix(render): coerce numeric and boolean echo params

### LINKED ISSUE #2722: numbers in `global_vars` seem to need quoting
### Description

Here “12345” seems to need quoting unlike regular YAML syntax
```yaml
global_vars:
  - name: test
    type: echo
    params:
      echo: "12345"
```

### Steps to reproduce

_No response_

### Expected behavior

Allow numbers unquoted in contexts like this as well

### Screenshots

_No response_

### Logs

_No response_

### Your environment

Espanso 2.3.0 on openSUSE Tumbleweed

### Self-service

VXNCXNX:
Fixes #2722.

## The problem

A numeric `global_vars` entry fails to render, with an error that says the parameter is missing when it is right there:

```yaml
global_vars:
  - name: myvar
    type: echo
    params:
      echo: 12345
```

The YAML loader is doing the right thing: an unquoted `12345` is a YAML integer, so `convert_value` produces `Value::Number(Number::Integer(12345))` (`espanso-config/src/matches/group/loader/yaml/util.rs:44`). But the echo extension only ever matched `Value::String`:

```rust
if let Some(Value::String(echo)) = params.get("echo") {
```

so anything else fell to the `else` and reported `missing 'echo' parameter`. The user has to know to quote the value, and the error points them nowhere near that.

I confirmed the three cases before touching anything:

```
number   => Error(missing 'echo' parameter)
bool     => Error(missing 'echo' parameter)
string   => Success(Single("12345"))
```

@smeech noted in the thread that this is intentional YAML behaviour and that coercion is what's actually wanted here, the way `replace:` values already get coerced. This PR does exactly that, for `echo` only — `date`'s `offset:` still needs a real number, so I left it alone.

## The fix

Match numbers and booleans too, formatting each explicitly:

- `Number::Integer` / `Number::Float` via `to_string()` — shortest round-trip, so `0.1` stays `0.1` and `1e20` expands in full with no precision loss
- `Bool` as `true` / `false`

`Null`, `Array` and `Object` still error. `Null` is what a bare `echo:` produces, and rendering it as `null` would be worse than failing; arrays and objects have no sensible single-string form, and formatting them via `Debug` would leak Rust syntax into the user's text.

`schemas/match.schema.json` is widened to `["string", "number", "boolean"]` in the same commit, otherwise the editor extensions would keep red-underlining exactly the config this makes valid.

One thing worth knowing: `Number::Float(1.0)` renders as `1`, not `1.0`. That's Rust's float formatting and it's pinned in the tests so it isn't a surprise later.

## Tests

Two new tests alongside the existing ones: `echo_coerces_scalars` covers integers (including negative), floats (`1.0` and `1.5`), and both booleans; `echo_rejects_non_scalars` pins that `Null`/`Array`/`Object` keep erroring.

```
cargo test -p espanso-render   => ok. 60 passed; 0 failed
cargo test -p espanso-config   => ok. 79 passed; 0 failed
cargo fmt --check -p espanso-render      => clean
cargo clippy -p espanso-render --all-targets -- --deny warnings  => exit 0
```

### DIFF
--- a/espanso-render/src/extension/echo.rs
+++ b/espanso-render/src/extension/echo.rs
@@ -17,7 +17,7 @@
  * along with espanso.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-use crate::{Extension, ExtensionOutput, ExtensionResult, Params, Value};
+use crate::{Extension, ExtensionOutput, ExtensionResult, Number, Params, Value};
 use thiserror::Error;
 
 pub struct EchoExtension {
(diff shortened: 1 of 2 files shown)


VXNCXNX:
Rebased onto current `dev`, no conflicts. `cargo test -p espanso-render` gives 60 passed and 0 failed on the new base, `cargo fmt --check` is clean, and the schema JSON still parses.

The push moved CI into `action_required`, so the PR page shows no signal until someone approves the run.

`+51/-6` across `espanso-render/src/extension/echo.rs` and `schemas/match.schema.json`. Happy to rebase again if `dev` moves.

AucaCoyan:
Thank you so much! this is something I desired to do, but fell down into the urgent task list

AucaCoyan (approved):

PR #2789, opened 2026-09-02
Add Wayland feature check for kdotool command

tebriel:
I got confused thinking I'd built the app with the wayland feature but it was just because it was linux. There's no reason to log about it if it wouldn't even be used.


AucaCoyan:
Oh, sweet! I didn't realized it was wayland only. Thanks!

tebriel:
> Oh, sweet! I didn't realized it was wayland only. Thanks!

Yeah KDE + Wayland only! Thanks.

Issue #2792, opened 2026-09-04
Shipped espanso.service uses WantedBy=default.target, racing the display server at boot (worker panics 101 until X/compositor is up)


HaleTom:
### Description

The shipped systemd user unit (`packaging/linux/usr/lib/systemd/user/espanso.service`, installed as `/usr/lib/systemd/user/espanso.service`) is hooked to `WantedBy=default.target`:

```ini
[Unit]
Description=espanso

[Service]
ExecStart=/usr/bin/espanso launcher
Restart=on-failure
RestartSec=3

[Install]
WantedBy=default.target
```

`default.target` is reached **before** the graphical session. On X11 systems espanso talks to Xorg/XWayland via `XOpenDisplay`, so when the unit starts at boot it races the display server. If X isn't up yet, the worker panics (`failed to initialize detector module: detection source initialization failed`, `espanso/src/cli/worker/engine/mod.rs:146`) → daemon exits → launcher panics (`espanso/src/cli/launcher/mod.rs:196`) → main process exits with status 101 → `Restart=on-failure` cycles the whole unit every 3 seconds until the display appears. Journal from one boot (4 attempts, ~507ms each):

```
14:27:30 Started espanso.
14:27:31 [worker(2564)] using X11Source
14:27:31 [worker(2564)] thread 'engine thread' panicked at 'failed to initialize detector module: detection source initialization failed': espanso/src/cli/worker/engine/mod.rs:146
14:27:31 systemd: Main process exited, code=exited, status=101/n/a
14:27:34 restart → same panic
14:27:38 restart → same panic
14:27:41 restart → OK (display finally up)
```

On AppImage installs this restart cycle has an extra cost: each failed start leaves a dead FUSE mount at `/tmp/.mount_espans*` (see #2791 for that mechanism).

The fix is one line in the unit: `WantedBy=graphical-session.target` (plus `After=graphical-session.target` in `[Unit]`), so the unit starts with — and after — the graphical session instead of racing it.

### Steps to reproduce

1. X11 system, espanso installed with the shipped `espanso.service` (as the .deb/AppImage packaging and Arch AUR packages do), `WantedBy=default.target`.
2. `systemctl --user enable espanso` and boot the machine.
3. If the display server takes more than ~a second to come up after `default.target`, the unit panics and restarts 2–4 times (`journalctl --user -u espanso` shows `status=101` cycles with `failed to initialize detector module`).

### Expected behavior

[...]

Issue #2791, opened 2026-09-03
AppImage edition leaks dead FUSE mount points when daemon/worker outlive a panicking launcher (inherited runtime fds defeat runtime teardown)


HaleTom:
### Description

When espanso (AppImage edition) starts during boot **before the X11 display is ready**, it panics and systemd restarts it. Each restart cycle leaves behind a **dead FUSE mount point** at `/tmp/.mount_espansXXXXXX` (`Transport endpoint is not connected`). Over time these accumulate — on my machine, 3 dead mounts from a single boot's restart cycle plus 1 healthy one from the current running instance.

The AppImage runtime (type2-runtime) already has a full teardown chain: a keepalive pipe whose read end is inherited by the exec'd app; when the app exits, the pipe breaks (EPIPE) → runtime's `write_pipe_thread` SIGTERMs the FUSE daemon → libfuse signal handler exits `fuse_session_loop` → `fuse_unmount`. This works correctly for **single-process** apps.

It breaks for espanso because espanso forks its lifetime into children with no fd hygiene: the daemon and worker **inherit the keepalive-pipe read end**, so when the launcher/main process exits (panic or normal), the pipe does not break — the FUSE daemon never gets the EPIPE→SIGTERM and lingers. Its mount then dies uncleanly (the FUSE daemon is later killed as part of cgroup cleanup) without running its unmount path.

Relevant code paths (v2.4.0):
- `espanso/src/cli/launcher/daemon.rs:26-29` — `launch_daemon` uses plain `Command::spawn()` + `wait()`
- `espanso/src/cli/daemon/mod.rs:241-251` (`spawn_worker`) and `espanso/src/cli/daemon/mod.rs:276-301` (`restart_worker`) — same pattern

Rust's `std::process::Command` inherits **all** parent fds by default. Inside an AppImage, the exec'd process carries the AppImage runtime's internal fds — notably the keepalive-pipe read end and the mount-dir fd (dup2'd to 1023 by the runtime) — and these propagate to the daemon and from there to the worker.

Evidence (journal timeline, one boot):
```
14:27:30 systemd[2215]: Started espanso.
14:27:31 status=101 (X11 not ready → panic)
14:27:34 restart → 101
14:27:38 restart → 101
14:27:41 restart → OK (ran all day)
```
→ 3 dead `/tmp/.mount_espans*` mounts from the 3 panicked starts; the 4th (healthy, single process tree alive all day) unmounted cleanly on stop. Same pattern on Aug 27.

[...]

PR #2793, opened 2026-09-04
fix(linux): start espanso with systemd graphical-session, not default.target

HaleTom:
### Issue for this PR

Closes #2792

### Type of change

- [x] Bug fix

### What does this PR do?

Updates the shipped Linux systemd unit (`espanso/src/res/linux/systemd.service`) so espanso is tied to the graphical session instead of racing the display server. Each change is commented inline in the unit:

```diff
 [Unit]
 Description=espanso
+# Order the start after the graphical session: the display server may not be
+# reachable yet while graphical-session.target is starting, which would make
+# the worker panic with "failed to initialize detector module" (exit status
+# 101) and cycle via Restart=on-failure until the display appeared.
+After=graphical-session.target
+# Stop espanso when the graphical session stops, so it does not outlive
+# logout as an orphaned daemon.
+PartOf=graphical-session.target
 
 [Service]
 ExecStart={{{espanso_path}}} launcher 
 Restart=on-failure
 RestartSec=3
 
 [Install]
+# Start with the graphical session instead of racing the display server.
-WantedBy=default.target
+WantedBy=graphical-session.target
```

- `WantedBy=graphical-session.target` (closes #2792): `default.target` is reached before the display server. On X11 systems the worker connects to the display at startup; starting too early makes it panic (`failed to initialize detector module: detection source initialization failed`, `espanso/src/cli/worker/engine/mod.rs:146` → exit status 101), and `Restart=on-failure` cycles the unit every 3 seconds until the display finally appears — 4 starts on one observed boot, 3 of them panics.
- `After=graphical-session.target`: explicit ordering so the start happens after the session target is reached, not concurrently with it.
- `PartOf=graphical-session.target`: stop espanso when the graphical session stops, so it does not outlive logout as an orphaned daemon.

Related: #2791 (AppImage FUSE mount leaks — the restart cycle above is what leaves dead `/tmp/.mount_espans*` mounts behind; this PR removes the trigger, #2791 removes the leak mechanism).

- Reproduced the failure mode on Arch (systemd 261): `journalctl --user -u espanso` shows the 3× `status=101` panic cycle at boot before the display is up, with the exact panic chain quoted in #2792.
- Verified the fixed ordering locally with a user-level override (`add-wants graphical-session.target` + `After=graphical-session.target` drop-in): the unit starts once, after the session, with no panics.
- `systemd-analyze --user verify` on the rendered unit (with `{{{espanso_path}}}` substituted): clean. This check caught an earlier iteration of this PR that placed `StartLimitIntervalSec`/`StartLimitBurst` in `[Service]` ("Unknown key ... ignoring") — both keys belong in `[Unit]`.

### Checklist

- [x] I have tested my changes locally

### DIFF
--- a/espanso/src/res/linux/systemd.service
+++ b/espanso/src/res/linux/systemd.service
@@ -1,10 +1,24 @@
 [Unit]
 Description=espanso
+# Order the start after the graphical session: the display server may not be
+# reachable yet when default.target is reached, which made the worker panic
+# with "failed to initialize detector module" (exit status 101) and cycle
+# via Restart=on-failure until the display appeared.
+After=graphical-session.target
+# Stop espanso when the graphical session stops, so it does not outlive
+# logout as an orphaned daemon.
+PartOf=graphical-session.target
 
 [Service]
 ExecStart={{{espanso_path}}} launcher 
 Restart=on-failure
 RestartSec=3
+# A slow boot can exceed the default restart budget (5 starts within 10s),
+# after which systemd marks the unit failed and gives up until manual reset.
+# Widen the budget so slow displays never permanently disable espanso.
+StartLimitIntervalSec=60
+StartLimitBurst=10
 
 [Install]
+# Start with the graphical session instead of racing the display server.
 WantedBy=graphical-session.target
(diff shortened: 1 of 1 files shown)


AucaCoyan:
Thank you!

AucaCoyan (approved):

PR #2797, opened 2026-09-07
fix(detect): report X11 hotkey registration failures with hotkey name

HaleTom:
### Related issue

Found while investigating #2791 (same X11/AppImage investigation; this PR implements the missing hotkey-name logging that came up there).

### Type of change

- [x] Bug fix

### What does this PR do?

On X11, `XGrabKey` reports `BadAccess` asynchronously through the error handler — never through its return value. `detect_register_hotkey` checked the return value (`res == BadAccess`), which is dead code, so a conflicting hotkey produced anonymous X11 error lines and the existing `unable to register hotkey` log never fired.

This PR makes `detect_register_hotkey` install a request-scoped error handler and call `XSync`, capturing the actual error code and the failing request opcode; `HotKeyResult`/`RawHotKeyResult` gain `error_code` and `error_request_code` fields. The Rust side now logs a warning naming the hotkey with the error and request translated to protocol names (e.g. `BadAccess` on `X_GrabKey`), and always inserts the mapping (XRecord-based detection does not depend on grab ownership, so a conflicted hotkey keeps working).

> **No behavior change.** In 2.4.1 the `res == BadAccess` check is dead code (`XGrabKey` reports `BadAccess` only through the error handler, never as a return value), so conflicted hotkeys already kept working through XRecord. This PR changes only the logging: the behavior existed before — it was just invisible.

**Before** — four anonymous lines, no hotkey name (espanso 2.4.1 AppImage, journal, 2026-09-02):

```
2026-09-02T14:27:41+07:00 svelte espanso[3811]: X11 Reported an error, code: 10, request_code: 33, minor_code: 0
2026-09-02T14:27:41+07:00 svelte espanso[3811]: X11 Reported an error, code: 10, request_code: 33, minor_code: 0
2026-09-02T14:27:41+07:00 svelte espanso[3811]: X11 Reported an error, code: 10, request_code: 33, minor_code: 0
2026-09-02T14:27:41+07:00 svelte espanso[3811]: X11 Reported an error, code: 10, request_code: 33, minor_code: 0
```

**After** — those four anonymous lines are replaced by a single warning naming the hotkey, the error, and the failing request (journal):

```
11:00:14 [worker(3968830)] [WARN] unable to register hotkey: ALT+SPACE (X11 error 10: BadAccess on X_GrabKey); the shortcut will still work but may also trigger that application
```

(The conflicting combo here is the built-in search shortcut, ALT+SPACE.)

### Error translation

The reported numbers are X protocol identifiers, translated programmatically on the Rust side:

- `error_code` → error name (`10` → `BadAccess`, `8` → `BadMatch`, …): the first 17 core-protocol error codes are mapped to their Xlib names; unknown codes fall back to the raw number.
- `error_request_code` → request name (`33` → `X_GrabKey`, …): the request opcode from `XErrorEvent`, so the message names the call that failed even if registration is later extended beyond `XGrabKey`. Unknown opcodes fall back to `request N`.
- `minor_code` is not captured: for core-protocol requests like `X_GrabKey` it is structurally always `0`, so it carries no information.

The conflict hint is attached only when the error is `BadAccess` — that is the one error meaning "another client owns the grab". Other errors mean a different problem, and the hint would mislead. The log level is `warn` rather than `error` because the mapping is always inserted and XRecord-based detection does not need grab ownership: the hotkey is degraded (it may not fire exclusively) but keeps working, so the condition is not fatal. (The original code paired `error!` with skipping the mapping — silently disabling a working shortcut.)

If maintainers would rather make exclusivity opt-in (e.g. a hotkey option like `exclusive = true` that drops the mapping on `BadAccess`), that would be a separate PR — `error_code` is already plumbed through the FFI. Happy to open one if there's interest.

- `cargo build -p espanso-detect` and `cargo build -p espanso` green; `clang-format --dry-run -Werror` clean.
- Live run on X11 with a conflicting ALT+SPACE grab: the WARN line above, followed immediately by `using X11ProxyInjector` / `using X11Clipboard` — the worker continued and the search shortcut still works, because the mapping is always inserted.

### Checklist

- [x] I have tested my changes locally

### DIFF
--- a/espanso-detect/src/x11/mod.rs
+++ b/espanso-detect/src/x11/mod.rs
@@ -223,10 +223,12 @@ impl Source for X11Source {
             let raw = convert_hotkey_to_raw(hk);
             if let Some(raw_hk) = raw {
                 let result = unsafe { detect_register_hotkey(handle, raw_hk, mod_indexes) };
-                // Any failing lock-mask variant sets error_code, so the warning
-                // can fire even when the plain combo registered fine; the hint
-                // covers the still-works case.
-                if result.error_code != 0 {
+                if result.success == 0 {
+                    error!("unable to register hotkey: {}", hk);
+                } else if result.error_code != 0 {
+                    // Any failing lock-mask variant sets error_code, so the warning
+                    // can fire even when the plain combo registered fine; the hint
+                    // covers the still-works case.
                     let hint = if result.error_code == BAD_ACCESS {
                         "; the shortcut will still work but may also trigger that application"
                     } else {
@@ -245,7 +247,10 @@ impl Source for X11Source {
                 }
                 // The mapping is inserted even when the grab failed: XRecord-based
                 // detection is unaffected by grab ownership, so the hotkey keeps working.
-                raw_hotkey_mapping.insert((result.key_code, result.state), hk.id);
+                // It is skipped only when no keycode resolved (success 0).
+                if result.success != 0 {
+                    raw_hotkey_mapping.insert((result.key_code, result.state), hk.id);
+                }
             } else {
                 error!("unable to generate raw hotkey mapping: {}", hk);
             }
--- a/espanso-detect/src/x11/mod.rs
+++ b/espanso-detect/src/x11/mod.rs
@@ -224,7 +224,10 @@ impl Source for X11Source {
             if let Some(raw_hk) = raw {
                 let result = unsafe { detect_register_hotkey(handle, raw_hk, mod_indexes) };
                 if result.success == 0 {
-                    error!("unable to register hotkey: {}", hk);
+                    error!(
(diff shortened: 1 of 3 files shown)


HaleTom:
Review bot verification: this branch was also reviewed by CodeRabbit and cubic on the mirror PR HaleTom/espanso#1 — both reported all findings addressed (CodeRabbit: merge risk low; cubic: "All reported issues were addressed"). Two bot-found issues (missing `success == 0` guard, generic error wording) were fixed in 94b4d78 and e7e8577. See https://github.com/HaleTom/espanso/pull/1 for the full review threads.

AucaCoyan (approved):
Thank you! 
It has a minor clippy fail, but I'll merge it and then I'll fix it on a new PR 👍🏼

